PR #12, opened 2021-03-11
UI form implemenation

giogio21:
Implementation of UI form and simple test

### DIFF
--- a/app/src/main/java/com/sdp13epfl2021/projmag/form.kt
+++ b/app/src/main/java/com/sdp13epfl2021/projmag/form.kt
@@ -1,0 +1,11 @@
+package com.sdp13epfl2021.projmag
+
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+
+class form : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_form)
+    }
+}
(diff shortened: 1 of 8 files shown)


app/src/main/res/layout/activity_form.xml
```diff
@@ -0,0 +1,105 @@
...
+        android:inputType="textPersonName"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/textView2"></EditText>
+
+    <EditText
+        android:id="@+id/EditTextName3"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
+        android:hint="Teacher"
```
arnauds5, line 48: You should use better name for id, instead of EditTextNameX, and avoid using hardcoded text (use resources instead).

app/src/main/java/com/sdp13epfl2021/projmag/form.kt
```diff
@@ -0,0 +1,11 @@
+package com.sdp13epfl2021.projmag
+
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+
+class form : AppCompatActivity() {
```
PaulCoral, line 6: class names should start with an uppercase letter (i.e. Form)
[Kotlin naming rules](https://kotlinlang.org/docs/coding-conventions.html#naming-rules)

PR #22, opened 2021-03-12
Feature/UI form

giogio21:


app/src/main/res/layout/activity_form.xml
```diff
@@ -0,0 +1,112 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginEnd="29dp"
+        android:text="@string/master_project"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toEndOf="@+id/CheckBoxSP"
+        app:layout_constraintTop_toBottomOf="@+id/nbOfParticipant"></CheckBox>
+
+    <EditText
+        android:id="@+id/projectDescription"
```
arnauds5: Because multiples elements will have a "projectDescription" for example, a better naming would be to contains the activity name (or at least to be uniquely identifiable).

PaulCoral:
Do you really have to merge `MainActivity` as it is not part of your Form (and `.idea/*`)?
Also I think that naming UI IDs would be better like this : [How do you name UI elements? (StackOverflow)](https://stackoverflow.com/a/7124990)

PaulCoral (approved):
Looks good for me. Clean and readable UI.

arnauds5:
Maybe hiding the submission button while we are writing could be great.
Except that, everything looks good to me.

arnauds5 (approved):
Maybe hiding the submission button while we are writing could be great.
Except that, everything looks good to me.

PR #11, opened 2021-03-10
Add Async Interface with FireStore for projects

PaulCoral:
An Interace with FireBase FireStore to add and get projects from it.

PR #23, opened 2021-03-13
First database of project and tags 

kaourintin:
=> project is an abstract project class that contains the error handling through a sealed class.
Then we have a concrete project class that inherits from this abstract class and that actually adds nothing, but it is better this way in case we need added flexibility.

It will still be necessary to add more flexibility on some things (like how many students can participate in the project etc later on)

tags implements (for now) a model where tags have no whitespaces and no upper cases. Hence, if we have machinelearning or Machine Learning or mAchIneLeArnIng only one tag will be stored in the database which is machinelearning

### DIFF
--- a/app/src/main/java/com/sdp13epfl2021/projStructure/AbstractProject.kt
+++ b/app/src/main/java/com/sdp13epfl2021/projStructure/AbstractProject.kt
@@ -1,6 +1,5 @@
 package com.sdp13epfl2021.projStructure
 
-import java.lang.IllegalStateException
 
 public abstract class AbstractProject(
         val projectName: String, val labName: String,
@@ -12,19 +11,21 @@ public abstract class AbstractProject(
     companion object{
         private final const val TAKEN: Boolean = true
         private final const val FREE: Boolean = false
-    }
 
+    }
+    sealed class results{
+        object userCommandSuccess : results()
+        object userErrorProjectTaken : results()
+    }
      private var isTaken = FREE;
 
-    fun showProject(): String {
-        return "owner : $projectOwner \nlab : $labName \ndescription : $projectDescription \n"
-    }
 
-    fun setTaken(){
+    fun setTaken() : results{
         if(isTaken == TAKEN){
-            throw IllegalStateException("project was already taken!")
+            return results.userErrorProjectTaken
         }
         isTaken = TAKEN;
+        return results.userCommandSuccess
     }
 
     fun setFree(){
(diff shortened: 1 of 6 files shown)


arnauds5:
As talked on discord, it will probably be better to have an immutable class for the projects.
Attribute naming is a little redundant : Project.projectName could be remplaced by Project.name
Adding an id attribute would be great. You can see some attributes on this https://github.com/asfoury/projmag/pull/11/commits/4f7ab79973ae53d6f76ef2700839edf2d4f504e0

asfoury (commented):
Everything looks good but I do not understand how the assigned students will be handled.
Also can you add your code in a package called model inside the projmag package like on the main branch?

app/src/main/java/com/sdp13epfl2021/projStructure/AbstractProject.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    }
+    sealed class results{
+        object userCommandSuccess : results()
+        object userErrorProjectTaken : results()
+    }
+     private var isTaken = FREE;
+
+
+    fun setTaken() : results{
```
PaulCoral, line 23: maybe you could use [kotlin getter and setter](https://kotlinlang.org/docs/properties.html#getters-and-setters), instead of `setTaken()` and `isTaken()`

app/src/main/java/com/sdp13epfl2021/projStructure/Project.kt
```diff
@@ -0,0 +1,11 @@
+package com.sdp13epfl2021.projStructure
+
+ public class Project(projectName: String, labName: String, responsibleName: String,
```
PaulCoral, line 3: could this be a data class ?

app/src/main/java/com/sdp13epfl2021/projStructure/AbstractProject.kt
```diff
@@ -0,0 +1,45 @@
+package com.sdp13epfl2021.projStructure
+
+
+public abstract class AbstractProject(
```
PaulCoral, line 4: could this be a data class ?

PaulCoral (changes requested):
Maybe use kotlin _getter/setter_ and _data class_

app/src/main/java/com/sdp13epfl2021/projStructure/AbstractProject.kt
kaourintin: Hi!
the kotlin data class has one main problem for my implementation : 
 the copy method we would want to use for immutability purposes does not enable us to return wrapped values and would force me to throw exeptions for error handling in the init. Exeptions are bad in kotlin because there is no way  to enforce exeption handling.

So I chose another design.

PR #29, opened 2021-03-15
Feature/cached projects database

### LINKED ISSUE #19: Add firebase listeners, to keep the local projects list up to date (anytime the firebase server is modified through adding or removing or modifying a project it is also updated on the phone)

arnauds5:
The new CachedProjectsDatabase is a ProjectsDatabase with one additional function : getAllProjects() that is synchronous.
I added to ProjectsDatabase interface, a way to add listeners to projects changes, that could be used instead of retrieving all projects every times.


kaourintin (approved):
We have to be careful of how the listeners are handled especially their deletion when projects are deleted by the user. 

Overall I am satisfied by the code and understand that the test coverage can't be perfect

PaulCoral (approved):
The coverage is now ok
Implementation is safe with synchronization
We are more in control with this cache than Firestore cache

PR #31, opened 2021-03-15
Feature/project database

kaourintin:
1)
=> database for projects which is immutable and contains two functions 
One that allows for the creation of a project from scratch
and another that allows for the partial copying of a project and allows the programmer to specify which fields change and give them new values.

Both functions return either a Success(project) or a Failure(reason:String) 

2)
=> database for tags which contains a mutable string. For now I allowed multiple instances to be created but will eventually modify it to a singleton class to make sure that there is only one tag list in each app on each phone.

The database stores tag in only one way => MachineLearning, machinelearning, Machine learning, machinel learning are all one unique tag which is machinelearning. no characters but normal letters are allowed in tags and will generate one of the errors of the sealed class.


app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBase.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+        else{
+            tags.add(cleanTag)
+            return InputResult.OK
+        }
+
+    }
+
+    //should I be making a defensive copy here?
+    fun getAllTags() : Set<String>{
```
PaulCoral, line 67: A copy would be preferred as we can cast it to mutable set again.
You can use `.toSet()` as shown [here](https://stackoverflow.com/a/49053542)

app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBase.kt
```diff
@@ -0,0 +1,71 @@
...
+        return tags
```
PaulCoral, line 68: Maybe safer with a defensive copy here
`tags.toSet()`

PaulCoral (approved):
See comment for defensive copy
`tags.toSet()`

asfoury (commented):
Everything looks good, something that is important to do before merging is to make ImmutableProject inherit from Parcelable so that we can pass Projects around between activities using Intents:
All you have to do is add this in the app: build.gradle:
plugins {
...

 id 'kotlin-parcelize' <- this line 
...
}

Then make ImmutableProject inherit from Parcelable and add `@Parcelize` before the class decleration.
Example:
image

It will show an error but you can ignore it its a bug in the IDE.

asfoury (approved):
Looks good, no need to make it Parcelizabe now may do that in the upcoming weeks.

PR #38, opened 2021-03-18
immutable project  has now exactly the same word struct as dummy

### LINKED ISSUE #24: Make project class immutable and add extra fields
As talked on discord, it will probably be better to have an immutable class for the projects.
Attribute naming is a little redundant : Project.projectName could be remplaced by Project.name
Adding an id attribute would be great. You can see some attributes on this https://github.com/asfoury/projmag/pull/11/commits/4f7ab79973ae53d6f76ef2700839edf2d4f504e0

_Originally posted by @arnauds5 in https://github.com/asfoury/projmag/issues/23#issuecomment-798674661_

kaourintin:
modified immutable project so that it's easier for other programmers to migrate from dummyProject to ImmutableProject by making the word structure exactly the same (same order, same names) as the dummy project.

Added tags and id to the immutable project. Tags dont implement the tag database yet

All the tests pass and there are no major code differences.

### DIFF
--- a/app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
+++ b/app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
@@ -4,50 +4,62 @@ sealed class Result<T>
 data class Success<T>(val value: T) : Result<T>()
 data class Failure<T>(val reason: String) : Result<T>()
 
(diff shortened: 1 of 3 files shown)


PaulCoral:
This branch is almost compatible, so I forked at #41 to implement the specific `FirebaseProjectDatabase` part

PR #39, opened 2021-03-18
implemented search function and ui, need to test

lpatel99:


PaulCoral (approved):
Looks, good very smooth implementation,
maybe close the keyboard when not focusing, the search bar

arnauds5 (approved):
Works fine, but there is a small bug where the some tags are not well displayed after using the search

PR #56, opened 2021-03-20
Feature/file database

### LINKED ISSUE #50: Update firebase interface to store videos (and files in general)

arnauds5:
This new database allow get/push/delete of file (can be used for videos/images/etc).
When a file is push with the firebase impl, it stores it at /users/{userUID}/{filname}, and give a full url.
When a file is get, it download from a url to a local folder using the same name.
Does anyone needs more specific functionallity and/or change on the FileDatabase interface ?
(for example : get metadata, give new name for get, allow to cancel download)
I'm writting test now.


lpatel99 (approved):
Looks good to me and works.

giogio21 (approved):
look good for me

PR #57, opened 2021-03-20
Added the button to the projects list that navigates to the form

asfoury:
The list of projects view now has a Floating Action Button that when clicked takes the user to the form activity.

### DIFF
--- a/app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
+++ b/app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
@@ -1,10 +1,14 @@
 package com.sdp13epfl2021.projmag.activities
 
+import android.content.Intent
 import android.os.Bundle
 import android.view.Menu
+import android.view.View
 import androidx.appcompat.app.AppCompatActivity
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
```diff
@@ -20,6 +24,14 @@ class ProjectsListActivity : AppCompatActivity() {
         itemAdapter = ItemAdapter(this, myDataset)
         recyclerView.adapter = itemAdapter
         recyclerView.setHasFixedSize(true)
+
+        // get the fab and make it go to the Form activity
+        val fab: View = findViewById(R.id.fab)
+        fab.setOnClickListener {
```
PaulCoral, line 30: Maybe would it be cleaner to set the `onClickListener` in the XML or UI editor, with the `onClick` option

PaulCoral (approved):
It is fine for me
Maybe just adding the listener in the UI editor, instead of putting it within the onCreate(...) method

kaourintin (approved):
Don't see anything to report in the code. Works great in the app and didn't find a way to crash it.

Might suggest adding an s to participant that's all.

PR #59, opened 2021-03-23
Feature/profile class for UI

kaourintin:
First draft for giovani to use, it's already been tested


lpatel99 (approved):
All good, I changed genderEnum to Gender to respect coding conventions.

giogio21 (approved):
look good for me

app/src/main/java/com/sdp13epfl2021/projmag/model/Gender.kt
```diff
@@ -0,0 +1,17 @@
...
+        },FEMALE{
+        override fun toString(): String {
+            return "Female"
+        }
+                },OTHER{
+        override fun toString(): String {
+            return "Other"
+        }
+                }
+}
```
arnauds5, line 17: This could be simplified with something like this
enum class Gender(private val text: String) {
    MALE("Male"), FEMALE("Female"), OTHER("Other");
    override fun toString(): String { return text }
}

PR #63, opened 2021-03-24
 UI profile

giogio21:


lpatel99 (approved):
I corrected a typo and made the profile button lead to the profile UI, all good otherwise.

PaulCoral (approved):
Seems good to me,
Maybe need some reformatting  and  coverage ,but the code is ok

PR #76, opened 2021-03-27
Feature/tags database

kaourintin:
Main changes : 
Phone now contains a singleton list of allowed tags
Projects now contain allowed tags from the singleton list
this list is thread safe and copy protected 
Tests were updated and change to accomodate this

Might need to add more tests for coverage?


app/src/main/java/com/sdp13epfl2021/projmag/model/Tag.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Class for tags, they are stored in lowercase and without spaces
+ *
+ * @property name : name of the tag
+ */
+data class Tag(var name : String) {
+    private var accepted : Boolean = false
+    init{
+        name = name.toLowerCase(Locale.ROOT).replace(" ","")
```
asfoury: I think you should also remove tabs and new lines as if I enter : "hello    world"(with a tab between hello and world) the replace function wont remove the tab, and we will have tags with tabs and new lines in the tag set.

asfoury (commented):
Everything looks good, just maybe look at my comment so we can make the tag set more robust

app/src/main/java/com/sdp13epfl2021/projmag/model/Tag.kt
kaourintin: Fixed, did a regex to hunt for all the kinds of ways we can add spaces instead of just spaces.

kaourintin: fixed with a regex that hunts for all the ways you can actually add spaces

asfoury: Nice! I will approve the PR

asfoury (approved):

lpatel99 (approved):
Tested extensively and the code looks good.

PR #79, opened 2021-03-28
Feature/local storage

arnauds5:
OfflineProjectsDatabase stores project's info in /{appDir}/projects/{projedId}/project.data.
Any others files (images/videos/etc), will be stored in the same directory.
The offline database give priority to the remoteDatabase if available (up to date).
The FileDatabase now considers every file push as new file (with random/unique name added).


app/src/androidTest/java/com/sdp13epfl2021/projmag/database/FakeDatabaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    fun add(project: ImmutableProject) {
+        projects = projects + project
+        listeners.forEach{ it -> it(ProjectChange(ProjectChange.Type.ADDED, project)) }
+    }
+
+    fun modify(project: ImmutableProject) {
+        projects.filter { p -> p.id != project.id }
+        projects = projects + project
+        listeners.forEach{ it -> it(ProjectChange(ProjectChange.Type.MODIFIED, project)) }
```
kaourintin: 1)Why are you not reaffecting the value to project in projects.filter
2) Why are you not checking that filter did find a previously existing project before adding the modified project (meaning if this was a project that did not exist in the projects list, then modify will do exactly the same as the add function, is this wanted behaviour?)

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/CachedProjectsDatabaseTest.kt
```diff
@@ -226,91 +226,4 @@ class CachedProjectsDatabaseTest {
...
-            id: ProjectId,
-            onSuccess: () -> Unit,
-            onFailure: (Exception) -> Unit
-        ) {
-            projects.find {p -> p.id == id}?.let { remove(it) }
-            onSuccess()
-        }
-
-        override fun addProjectsChangeListener(changeListener: (ProjectChange) -> Unit) {
-            listeners = listeners + changeListener
```
kaourintin, line 308: Explain more explicitly by commenting what listeners are for and what they are associated to (because I thought there was one listener per project and it took me a while of reading your code to understand)

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/FakeDatabaseTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+import com.sdp13epfl2021.projmag.model.ImmutableProject
+import java.util.*
+
+
+class FakeDatabaseTest(projectsBeginning: List<ImmutableProject>) : ProjectsDatabase {
+
+    var projects: List<ImmutableProject> = projectsBeginning
+    var listeners: List<((ProjectChange) -> Unit)> = emptyList()
+    var nextId: Int = 90000
```
kaourintin, line 11: ugly solution but for now this is fine

app/src/main/java/com/sdp13epfl2021/projmag/database/OfflineProjectDatabase.kt
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @Synchronized
+    private fun loadProjects() {
+        projectsFiles = projectsDir
+            .listFiles()!!
+            .map { child -> child.name to File(child, PROJECT_DATA_FILE) }
+            .filter { (id, data) -> id.matches(ID_PATTERN) && data.exists() && data.isFile }
+            .toMap()
+    }
```
kaourintin, line 47: please comment long lambdas with natural language so it's a bit easier to understand (even though in this case it's easy to understand)

kaourintin (changes requested):

kaourintin:
all code climate issues can be safely ignored and we might want to bump up code climate code requirements to something like 35 lines per function and change the "cognitive limits"

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/FakeDatabaseTest.kt
arnauds5: 1) It was a mistake. Thanks for reporting that error.
2) The modify method (inside the test class) is used to trigger a MODIFIED change, and yes the effect on projects is the same as the add.

kaourintin: But if we have :

Database(project(A), project(B))

and we call modify on this database with a project(C)
the database becomes Database(A,B,C), adding a project, but the listeners receive a MODIFIED notice instead of an ADD notice ? no?

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/CachedProjectsDatabaseTest.kt
arnauds5: Listeners on ProjectsDatabase are triggers at every projects change (i.e. on new project, on deleted project or on modified project).
The listeners are not associated to a particular project but on the whole collection of projects.

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/FakeDatabaseTest.kt
arnauds5: This class is only used in tests, as a fake db to provide a ProjectDatabase.
The nextId in firebase is random and unique, here I only used a counter because only a few known projects are added.

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/FakeDatabaseTest.kt
arnauds5: Yes that's correct. And this can be the case in the real ProjectDatabase, for example : 
We have p1, p2 in our phone, we lost connection, p3 is added to firebase, we reconnect, p3 is modified => we will receive modified.

kaourintin (approved):

kaourintin (approved):

kaourintin (approved):

app/src/main/java/com/sdp13epfl2021/projmag/database/Utils.kt
```diff
@@ -1,9 +1,43 @@
...
+                        ),
+                        File(context.filesDir, "projects")
+                    )
+                )
+            fileDatabase = FirebaseFileDatabase(Firebase.storage, Firebase.auth)
+            userDatabase = UserDataFirebase(firestore, Firebase.auth)
+            initalized = true
+        }
+    }
 }
```
PaulCoral: Maybe it would be safer, and would also make more sense, to have a `class` instead of an `object` if we have an `init` method.

Imagine we keep the `context` between activities, this would not be expected.
However, if we instantiate an object an object with `context`, it would stay within the `activity` and not share the `context` with another one.
However, you can still put static stuff in a `companion object`

PaulCoral (changes requested):
Utils could become an class, because it has to be initialized. Now it can share activity context and maybe we should avoid this behavior.

app/src/main/java/com/sdp13epfl2021/projmag/database/Utils.kt
arnauds5: I changed the Utils object to a class and its constructor ask for a context.

arnauds5:
The project view list is now always up to date. Any modification of projects on firebase is directly updated without any need to refresh manually.

PaulCoral (approved):
After the fixes, it is okay now

PR #80, opened 2021-03-29
Removed intent filters from the ProfilePageActivity tag

asfoury:


kaourintin (approved):
Bug fix accepted

lpatel99 (approved):
Tested, the problem is indeed fixed.

PR #81, opened 2021-03-30
Feature/project info view with videos

### LINKED ISSUE #74: Install the video in local storage using the firebase functions

### LINKED ISSUE #77: Change the project information view to support images and videos

### LINKED ISSUE #94: The app crashed if an invalid project is get from firebase/local. (for example missing a field or wrong type)
https://github.com/asfoury/projmag/blob/912ac16f4faaa326aaf348ab886ca5dcb316f957/app/src/main/java/com/sdp13epfl2021/projmag/database/FirebaseProjectsDatabase.kt#L31
https://github.com/asfoury/projmag/blob/912ac16f4faaa326aaf348ab886ca5dcb316f957/app/src/main/java/com/sdp13epfl2021/projmag/database/OfflineProjectDatabase.kt#L84

arnauds5:
The description of project is now handled as basic html.
Images/videos are download from firebase storage to local storage.
Images stay inside the description, videos are added at the end with a controller to select videos.

Some examples ("There are videos in this project" or "Implement a fast driver for a 100 Gb/s network card")
image

android_23_oneplus2 android 23


arnauds5:
I had some tests that worked fine before, but they can't work anymore because I can't modify database for testing.
I need to either find a way to replace database during testing, or to use real projects but the tests won't be reliable.

arnauds5:
As a temporary way to handle video url, was with \<video>url\</video> tags inside the description.  
This will probably be changed to a list of url in ImmutableProject in another PR.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -1,37 +1,299 @@
...
 class ProjectInformationActivity : AppCompatActivity() {
+
+    private lateinit var fileDB: FileDatabase
+    private lateinit var videoView: VideoView
+    private var videosUris: List<Uri> = emptyList()
+    private var current: Int = -1
+
+    @Synchronized
+    private fun addVideo(videoUri: Uri, videoView: VideoView) {
+        videosUris = videosUris + videoUri
```
PaulCoral: Maybe a `MutableList`

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -1,37 +1,299 @@
...
+        controller.setPrevNextListeners({ next ->
+            readVideo(current + 1, videoView)
+        }, { prev ->
+            readVideo(current - 1, videoView)
+        })
+
+        // play the next video when the video is finished
+        videoView.setOnCompletionListener {
+            readVideo(current + 1, videoView)
+        }
```
PaulCoral: A `readNextVideo()` and `readPrevVideo()` would maybe be cleaner

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -1,37 +1,299 @@
...
+                            Toast.LENGTH_SHORT
+                        ).show()
+                    }
+                }
+            }, {
+                Toast.makeText(
+                    context,
+                    "An error occurred while downloading image.",
+                    Toast.LENGTH_SHORT
+                ).show()
```
PaulCoral: Maybe you could reduce this code section by creating a `showToast()` method

PaulCoral (commented):
**General remark** : when you document your method, you should use `/** doc */` instead of one line comment `//`

kaourintin (approved):
=> might need to add a textView explaining to the user that multiple videos can be contained inside the video slot
=> project building is not using the provided safe functions from the project class to build. Might be  a good task for next week.
=> Text in small char groups move to the left when too big which is very ugly. Should be solved when proper project creator functions are used (does not concern this PR)
=> code climate is being a baby and complains about functions that are simple and would be annoying and useless to make with less lines

that's a go from me

arnauds5:
The projects are now correctly ignored if they can't be loaded.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -1,37 +1,284 @@
...
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_project_information)
 
+        fileDB = Utils(this).fileDatabase
+
         // get all the text views that will be set
         val title = findViewById<TextView>(R.id.info_project_title)
```
lpatel99, line 76: could possibly rename values to be consistent.

lpatel99 (approved):
Tested, works well.

PR #96, opened 2021-04-13
Add some fix and tests

### LINKED ISSUE #99: TagBaseManager(TagsBase) is not Thread safe
In TagBaseManager, it is mentioned that TagsBase should be thread safe, but it is not.
Using the following test, it sometimes failed.
```
@Test
fun isTagBaseManagerThreadSafe() {
    val tagBase = TagsBaseManager()
    val initialSize = tagBase.getAllTags().size
    val threads: MutableList<Thread> = ArrayList()
    for (i in 1..40) {
        threads.add(Thread {
            val resulta = tagBase.addTag(Tag("a".repeat(i)))
            TestCase.assertEquals(TagsBaseManager.InputResult.OK, resulta)
        })
    }
    threads.forEach(Thread::start)
    threads.forEach(Thread::join)
    assertEquals(40 + initialSize, tagBase.getAllTags().size)
}
```

arnauds5:
Add videoURI to the buildCopy
Fix thread safety of TagsBaseManager
delete local projects if invalid
Add some tests (one is disabled because of cirrus)


lpatel99 (approved):
All these waiting times might be problematic and significantly slow down testing time. Regardless great tests and contributions.

app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBaseManager.kt
```diff
@@ -59,13 +59,15 @@ import java.util.regex.Pattern
...
-            return InputResult.OK
+        synchronized(tags) {
+            //various other checks :
+            if(tags.contains(tag)){
+                return InputResult.AlreadyExists
+            }
+            else{
+                tags.add(tag)
+                return InputResult.OK
+            }
```
kaourintin, line 70: thanks

kaourintin (approved):

PR #82, opened 2021-04-01
add video to cv

giogio21:

PR #97, opened 2021-04-14
Feature/profile classe teacher student

giogio21:


arnauds5:
It seems that a lot of code is dupplicated here.
If the only differences are not having a sciper for teacher, and to differentiate between student/teacher, maybe it could be done by : 
- Having a nullable sciper
- Adding an attribute with an enum of Student/Teacher/etc

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProfile.kt
```diff
@@ -21,18 +21,19 @@ class ImmutableProfile private constructor(val lastName : String, val firstName
...
         fun build(lastName : String, firstName : String,
-                  age : Int, gender : Gender, sciper : Int, phoneNumber :String) : Result<ImmutableProfile> {
+                  age : Int, gender : Gender, sciper : Int?, phoneNumber :String, role : Role) : Result<ImmutableProfile> {
             return when{
                 lastName.length > MAX_LAST_NAME_SIZE ->  Failure("last name is more than $MAX_LAST_NAME_SIZE characters")
                 firstName.length > MAX_FIRST_NAME_SIZE -> Failure("first name is more than $MAX_FIRST_NAME_SIZE characters")
                 age > MAX_AGE -> Failure("age is more than $MAX_AGE")
                 age <  MIN_AGE -> Failure("age is less than $MIN_AGE")
-                sciper < 0 -> Failure("sciper can't be a negative value")
+                (sciper!! < 0) -> Failure("sciper can't be a negative value")
```
arnauds5: If we allow null sciper, this check will throw a NullPointerException for null value
```suggestion
                (sciper != null && sciper < 0) -> Failure("sciper can't be a negative value")
```

arnauds5 (approved):
Looks good to me.
Maybe adding the video part, only in the other PR would have been better, but seems ok.

PaulCoral:
Same as in #82 

- [1](https://github.com/asfoury/projmag/pull/82#discussion_r614090406)
- [2](https://github.com/asfoury/projmag/pull/82#discussion_r614093646)

app/src/main/java/com/sdp13epfl2021/projmag/model/Role.kt
```diff
@@ -0,0 +1,17 @@
...
+        }
+           },TEACHER{
+        override fun toString(): String {
+            return "Teacher"
+        }
+    },OTHER{
+        override fun toString(): String {
+            return "Other"
+        }
+    }
```
PaulCoral, line 16: [Suggestion] this is a shorter and clearer syntax, but also ok to keep it
```kotlin
override fun toString(): String = "Other"
```

app/src/main/AndroidManifest.xml
```diff
@@ -13,9 +22,15 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.Projmag">
+
+        <activity android:name=".VideoCv">
+
+        </activity>
+
```
PaulCoral: **There's no VideoCv Activity**

PaulCoral (approved):
Looks good to me
**It would be better to remove `.VideoCv` from `AndroidManifest`**

PR #101, opened 2021-04-14
Add subtitle to project video

### LINKED ISSUE #89: Allow user to add subtitles to project video

PaulCoral:
PR associated to issue #89 (closes #89 when merged)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -5,6 +5,7 @@
     <uses-permission
         android:name="android.permission.READ_EXTERNAL_STORAGE"
         android:required="true" />
+
     <application
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
@@ -12,6 +13,7 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.Projmag">
+        <activity android:name=".video.VideoSubtitlingActivity"></activity>
         <activity android:name=".ProfilePageActivity" />
         <activity
             android:name=".curriculumvitae.CVCreationActivity"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,7 +13,9 @@
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:theme="@style/Theme.Projmag">
-        <activity android:name=".video.VideoSubtitlingActivity"></activity>
+        <activity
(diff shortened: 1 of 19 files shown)


PaulCoral:
With the current implementation, we can :

- Add subtitles to the provided video. Currently, only English caption is supported. Maybe we will add other language in the future.
- Watch a feed back video, with the subtitling we are adding in the `VideoSubtitlingActivity`
- Send the subtitles to firebase in a video_metadata collection, which link a pair (language_id,subtitles) with the video uri

PaulCoral:
The diff coverage is low because git detects more change than I really did, and these changes already had a low coverage.
that's why the total coverage is unchanged.
So please _ignore_ the coverage the **-4%** in diff-coverage

arnauds5 (approved):
Looks good to me.
Just to be sure, is it normal that there is no subtitle on the preview in Form.kt but only on VideoSubtitling ? (maybe it could be usefull to do a final check before sending a project)

app/src/main/AndroidManifest.xml
```diff
@@ -20,7 +24,9 @@
             android:name=".activities.ProjectInformationActivity"
             android:parentActivityName=".activities.ProjectsListActivity" />
         <activity android:name=".SignInActivity" />
-        <activity android:name=".Form" />
+        <activity
+            android:name=".Form"
+            android:screenOrientation="portrait" />
```
kaourintin, line 29: Error from Cirrus :
Expecting `android:screenOrientation="unspecified"` or `"fullSensor"` for this activity so the user can use the application in any orientation and provide a great experience on Chrome OS device

To me it seems important, exept if we want to keep the app in only portrait mode

PaulCoral:
@arnauds5 
> Looks good to me.
> Just to be sure, is it normal that there is no subtitle on the preview in Form.kt but only on VideoSubtitling ? (maybe it could be usefull to do a final check before sending a project)

Yes good suggestion ! It will be added in the next commits

kaourintin (approved):
Worked on my phone but not on my cmpter. Works on the computer of others so I will retry after reinstalling my emulator. Code looks good, exept for the cirrus warning that app is stuck in portrait mode

PaulCoral:
Closes #87 
Closes #88 
Closes #85

arnauds5 (approved):

app/src/main/AndroidManifest.xml
PaulCoral: Thanks for your review.
I force the portrait orientation as other orientation provide a poor user experience, and avoid some visual glitch, slowdowns when changing accidentally orientation.

PR #118, opened 2021-04-19
Change Utils to a singleton

### LINKED ISSUE #117: Every instance of Utils use the same database
Because Utils loads all projects at it's creation, a lot of unnecessary network/disk operations are done.
Because All Utils instances can modify local files, it could raise concurrent files modifications.

arnauds5:
Using a singleton, it avoid the mentionned problems.
Only a single instance of database could exist with this change.
No leak can happen because context is not stored in Utils anymore.
If we call Utils.getInstance before tests with dummy databases, we can now mock the entire databases.

### DIFF
--- a/app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectCreationActivity.kt
+++ b/app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectCreationActivity.kt
@@ -190,7 +190,7 @@ class Form : AppCompatActivity() {
          */
         private fun submit(view: View) = Firebase.auth.uid?.let {
             setSubmitButtonEnabled(false) // disable submit, as there is a long time uploading video
-            val utils = Utils(this)
+            val utils = Utils.getInstance(this)
             ProjectUploader(
                 utils.projectsDatabase,
                 utils.fileDatabase,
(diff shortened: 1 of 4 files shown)


lpatel99 (approved):
good initiative! all good

arnauds5:
I change the constructor of Utils & the getInstance. (Firebase class was not mockable)
Now every databases are optional in Utils.getInstance.
If you need to have constant database in your tests (and not rely on firebase), you can pass a fake database to the getInstance in a @Before function.

PaulCoral (commented):
Wouldn't these changes share `Context` between activities ?
Maybe everything could be static (object or companion object), except instances relying on a `Context`, where some get function would ask for it?

arnauds5:
> Wouldn't these changes share `Context` between activities ?

No, context is only use in the constructor but it is not stored anymore (val removed).

PaulCoral:
> > Wouldn't these changes share `Context` between activities ?
> 
> No, context is only use in the constructor but it is not stored anymore (val removed).

So only the `applicationContext` is saved and already shared across activities ???

arnauds5:
> > > Wouldn't these changes share `Context` between activities ?
> > 
> > 
> > No, context is only use in the constructor but it is not stored anymore (val removed).
> 
> So only the `applicationContext` is saved and already shared across activities ???

I removed the context from the constructor, now no context is used in the Utils instance, only on getInstance to initialize databases if necessary.

PaulCoral (approved):
Looks fine after @arnauds5 answered my questions

PR #124, opened 2021-04-20
Feature/section UI

### LINKED ISSUE #106: Implement a class (like tags) to say to which section(s) the project belongs to

### LINKED ISSUE #110: Add the section identifier to the project itself and modify the UI accordingly

### LINKED ISSUE #111: Add the section identifier to the UI of the project creation

kaourintin:
=>model modified so as to implement sections
=>Section creation UI in builder is finished and communicates with the builder
=> for now sections are in the tag groupChip of projects 

TODO: 
=> testing on the section manager and the section UI (done)
=>fix that when exiting and reentering the Tags activity and Section activity  the previous state is not shown 
=>Try to improve the UI ( the box is green instead of the text)  (done)


app/src/androidTest/java/com/sdp13epfl2021/projmag/database/OfflineProjectsDatabaseTest.kt
```diff
@@ -16,9 +19,60 @@ class OfflineProjectsDatabaseTest {
     private val projectsDir = Files.createTempDirectory("offline_test").toFile()
     private val unreadableDir = Files.createTempDirectory("offline_test_invalid").toFile()
 
-    private val p1 = ImmutableProject("12345","What fraction of Google searches are answered by Wikipedia?","DLAB","Robert West","TA1",1, listOf<String>(),false,true, listOf("data analysis","large datasets","database","systems","database","systems"),false,"Description of project1")
-    private val p2 = ImmutableProject("11111","Real-time reconstruction of deformable objects","CVLAB","Teacher2","TA2",1, listOf<String>(),false,true, listOf("Computer Vision","ML"),false,"Description of project2")
-    private val p3 = ImmutableProject("00000","Implement a fast driver for a 100 Gb/s network card","DSLAB","Teacher5","TA5",3, listOf<String>(),false,true, listOf("Low Level","Networking","Driver"),false,"Description of project5")
+    private val tagManager = TagsBaseManager()
+    private val sectionManager = SectionBaseManager()
+    private val tagList : List<String> = tagManager.tagsListToStringList(tagManager.getAllTags())
```
arnauds5: Using the tagsBaseManager could lead to tests failure, in particular because it has some default tags inside.
For example if there is not enough tags, tagList[3] will not be valid.

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fun isListValid(listToBeChecked : List<String>) : Boolean{
+        for(section in listToBeChecked){
+            if(!sectionsList.contains(section)){
+                return false
+            }
+        }
+        return true
+
```
arnauds5: Could be simplified by this
```suggestion
        listToBeChecked.all(sectionsList::contains)
```

app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBaseManager.kt
```diff
@@ -70,7 +70,7 @@ import java.util.regex.Pattern
                 return InputResult.AlreadyExists
             }
             else{
-                tags.add(tag)
+                tags.add(tag.copy())
```
arnauds5, line 75: This is not protecting enough, because we can get all Tags, even if we copy when adding, they can be modified after.

arnauds5 (changes requested):

app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBaseManager.kt
```diff
@@ -91,4 +91,13 @@ import java.util.regex.Pattern
...
      }
 
+     fun isListOfStringsValidTags(listTags : List<String>) : Boolean{
+         for(tag in listTags){
+             if(!tags.contains(Tag(tag))){
+                 return false
+             }
+         }
+         return true
+     }
```
arnauds5, line 100: This could be simplified by 
```suggestion
         listTags.map(::Tag).all(tags::contains)
```

app/src/test/java/com/sdp13epfl2021/projmag/model/ImmutableProjectTest.kt
```diff
@@ -125,33 +134,20 @@ class ImmutableProjectTest {
...
-                id = default_id
+                allowedSections = get("allowedSections") as List<String>
             )
         }
 
         assertEquals(project, projectRebuilt)
     }
 
-    @Test
+    @Test @Ignore
```
arnauds5: Is there a reason to ignore this test ?

app/src/main/java/com/sdp13epfl2021/projmag/adapter/ProjectAdapter.kt
```diff
@@ -104,6 +104,12 @@ class ProjectAdapter(private val context: Context, private val utils: Utils, pri
                 holder.chipGroupView.addView(chipView)
             }
 
+            for(section in project.allowedSections){
+                val chipView: Chip = Chip(context)
+                chipView.text = section
+                holder.chipGroupView.addView(chipView)
+            }
```
arnauds5: The tags and sections can't be differentiated. Maybe separating them or changing color of one type could be better,

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/OfflineProjectsDatabaseTest.kt
kaourintin: For now we assume that the tagDatabase contains at least three tags because we are the only ones with access to it

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
kaourintin: Begone, odersky fan, we do it the imperative way like it has been done for thousands of years before us

app/src/test/java/com/sdp13epfl2021/projmag/model/ImmutableProjectTest.kt
kaourintin: Thank you for taking the time to check out the test and solve the issue. This was un ignored

app/src/main/java/com/sdp13epfl2021/projmag/activities/SectionSelectionActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private fun handleListeningOnElementsOfSectionRecyclerView(sectionRecyclerView: RecyclerView, sectionDataset:List<String>){
+        sectionRecyclerView.addOnItemTouchListener(
+            MakeRecyclerListItemsClickListenable(
+                this,
+                sectionRecyclerView,
+                object : MakeRecyclerListItemsClickListenable.OnItemClickListener {
+
+                    override fun onItemClick(view: View, position: Int) {
+                        val holder = sectionRecyclerView.findViewHolderForLayoutPosition(position) as SectionAdapter.SectionViewHolder
```
arnauds5, line 53: It would be easier (and less lines of code) to just add a click listener in the SectionAdpater (same as ProjectAdapter),
and pass the already choosen section lists to the adapter constructor.

app/src/main/java/com/sdp13epfl2021/projmag/activities/SectionSelectionActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+        )
+    }
+
+    private fun saveButtonHandling(saveButton: Button){
+        saveButton.setOnClickListener{
+            val returnIntent = Intent()
+            val sections  = chosenSections.toTypedArray()
+
+
+            returnIntent.putExtra("sectionsList",  sections as Serializable)
```
arnauds5: Using a constant instead of a String literal could be better and could avoid typo on different files.
In MainActivityCompanion there are already a few const for intent name.

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
```diff
@@ -80,6 +82,9 @@ data class ImmutableProject constructor(
                     "there are ${assigned.size} " +
                             "students currently assigned but only $nbParticipant allowed to work for the project"
                 )
+                //!sectionsManager.isListValid(allowedSections) -> Failure("a section in the section list doesn't exist $allowedSections")
+                 //!tagsManager.isListOfStringsValidTags(tags) -> Failure("a tag in the tag list doesn't exist $tags")
+
```
arnauds5, line 87: Is this code commented because it breaks the projects already on Firebase ?

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
arnauds5: **!**(I aggree with you)

app/src/main/java/com/sdp13epfl2021/projmag/model/TagsBaseManager.kt
arnauds5: Same reason as the other one line / multilines ?

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
arnauds5: !(I aggree with you)

arnauds5:
There are still conflits with mains.

app/src/main/AndroidManifest.xml
```diff
@@ -28,23 +23,22 @@
...
         android:theme="@style/Theme.Projmag">
 
+        <activity android:name=".activities.SectionSelectionActivity"></activity>
+        <activity android:name=".curriculumvitae.VideoCv"></activity>
+        <activity android:name=".activities.TagsSelectorActivity" />
+
 
-        <activity android:name=".curriculumvitae.VideoCv"
-            android:parentActivityName=".activities.ProfilePageActivity">
-        </activity>
```
arnauds5: Is this normal the the parent activity is removed here ?

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
```diff
@@ -29,11 +27,13 @@ data class ImmutableProject constructor(
...
     val description: String,
-    val videoURI: List<String> = listOf()
+    val videoURI: List<String> = listOf(),
+    val allowedSections: List<String> = listOf()
 ) : Parcelable {
     companion object {
-
-        object FieldNames {
+        private val sectionsManager = SectionBaseManager()
+        private val tagsManager = TagsBaseManager()
```
arnauds5: Is there a reason to keep these managers here ?
They are private and never used.

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
```diff
@@ -0,0 +1,25 @@
...
+        return sectionsList
+    }
+
+    fun isListValid(listToBeChecked : List<String>) : Boolean{
+        for(section in listToBeChecked){
+            if(!sectionsList.contains(section)){
+                return false
+            }
+        }
+        return true
```
arnauds5: If the manager are only used to access the private list, maybe putting the function in the companion object and remove the the instanciation of Manager could simplify the code.

app/src/main/java/com/sdp13epfl2021/projmag/adapter/ProjectAdapter.kt
```diff
@@ -140,6 +141,24 @@ class ProjectAdapter(private val context: Context, private val utils: Utils, pri
...
+            chipView.chipBackgroundColor = ColorStateList.valueOf(ContextCompat.getColor(context, R.color.light_green))
+            holder.chipGroupView.addView(chipView)
+        }
+
+        for(section in project.allowedSections){
+            val chipView: Chip = Chip(context)
+            chipView.text = section
+            chipView.chipBackgroundColor = ColorStateList.valueOf(ContextCompat.getColor(context, R.color.teal_700))
+            holder.chipGroupView.addView(chipView)
+        }
```
arnauds5: The code seems a bit dupplicate, maybe adding a function like that would avoid it.
private fun addChips(textsList: List<String>, color: ...) {...}

app/src/main/java/com/sdp13epfl2021/projmag/adapter/ProjectAdapter.kt
```diff
@@ -140,6 +141,23 @@ class ProjectAdapter(private val context: Context, private val utils: Utils, pri
...
+        val green = ColorStateList.valueOf(ContextCompat.getColor(context, R.color.light_green))
+        val teal = ColorStateList.valueOf(ContextCompat.getColor(context, R.color.teal_700))
+        chipAdding(project.tags, holder, green)
+        chipAdding(project.allowedSections, holder,teal)
+    }
+
+    private fun chipAdding(list:List<Any>, holder:ProjectViewHolder, color:ColorStateList){
+        for(text in list){
+            val chipView: Chip = Chip(context)
+            chipView.text = text as CharSequence
```
arnauds5: Unsafe cast can be dangerous in general.
```suggestion
    private fun chipAdding(list: List<String>, holder: ProjectViewHolder, color: ColorStateList) {
        for (text in list) {
            val chipView: Chip = Chip(context)
            chipView.text = text
```

arnauds5:
Is it normal that the section/video button overlap ? (and/or buttons are not centred)
image

app/src/main/java/com/sdp13epfl2021/projmag/activities/SectionSelectionActivity.kt
kaourintin: Fixed in one of the last commits

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
kaourintin: Yes, for now we don't check this because most projects on firebase do not have the right tags and sections. This is a feature and not a bug

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
```diff
@@ -29,11 +26,11 @@ data class ImmutableProject constructor(
     val tags: List<String>,
     val isTaken: Boolean,
     val description: String,
-    val videoURI: List<String> = listOf()
+    val videoURI: List<String> = listOf(),
+    val allowedSections: List<String> = listOf()
```
arnauds5: This is not consistent with build(...) that has no default empty list for sections, was this the expected behaviour ?

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
```diff
@@ -159,7 +166,7 @@ data class ImmutableProject constructor(
      * @param isTaken : Is this project already taken
      * @param description : description of the project
      */
-    fun buildCopy(
+    fun rebuild(
```
arnauds5: Changing the name to rebuild could introduce some misunderstandings because it does not rebuild the project (immutable), but does create a copy with some new values.

app/src/main/java/com/sdp13epfl2021/projmag/model/SectionBaseManager.kt
```diff
@@ -0,0 +1,22 @@
...
+    companion object{
+        private val sectionsList = listOf<String>(
+            "Math", "Chemistry","Physics","CS", "Systems Communication","EE","Mechanical","Microtechnics"
+            ,"Material","Life Sciences","Architecture", "Civil","Environmental"
+        )
+
+        fun isListValid(listToBeChecked : List<String>) : Boolean = listToBeChecked.all(sectionsList::contains)
+
+        fun sectionList(): List<String> {
+            return sectionsList.toList() //safe copy
```
arnauds5: The sectionsList was already an unmodiable and safe List. (no need of .toList() in this case)

app/src/main/res/layout/list_project.xml
```diff
@@ -33,10 +33,11 @@
             android:padding="16dp"
             android:textAppearance="?attr/textAppearanceHeadline6" />
         <com.google.android.material.chip.ChipGroup
-            android:id="@+id/chip_group"
+            android:id="@+id/tags_chip_group"
```
arnauds5: Changing name could lead to misunderstanding because it is not longer only for tags, but also for sections.

app/src/main/java/com/sdp13epfl2021/projmag/model/Tag.kt
```diff
@@ -11,7 +11,7 @@ data class Tag(var name: String) {
     private var accepted : Boolean = false
     init{
 
-        name = name.toLowerCase(Locale.ROOT).replace("\\s+".toRegex(), "")
+        name = name.replace("\\s+".toRegex(), "")
```
arnauds5, line 14: It was already accepted in main, but is it really correct to remove space ?
For example the tag "Machine Learning" can't exist because it will be modified to "MachineLearning" with this replace.

arnauds5 (dismissed):

app/src/main/java/com/sdp13epfl2021/projmag/activities/TagsSelectorActivity.kt
```diff
@@ -54,25 +70,36 @@ class TagsSelectorActivity : AppCompatActivity() {
...
     }
 
+    /*
+    private fun colorPreviouslySelectedSections(previouslySelectedSections : List<String>){
+        for(section in previouslySelectedSections){
+
+            val holder = tagRecyclerView.findViewHolderForLayoutPosition(position)
+            holder.textView.setTextColor(Color.GREEN)
+        }
+    }*/
```
arnauds5: This code is not used and could be removed ?

PaulCoral (commented):
For me it is fine, however as it is a huge code change and I might have missed thing, I think it would be better to have **more than two reviews on this PR**

@arnauds5 has many good point, I do not see anything more to say

**We should definitely meet together for this PR**

By testing, I didn't noticed major issues

PaulCoral (approved):

arnauds5 (approved):
.

PR #126, opened 2021-04-22
Firebase connected to profile view

### LINKED ISSUE #112: Link the profile Information and activity to firebase storage

asfoury:
Draft request because still need to add tests and refactor the code


app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+     * @return document associated to the user
+     */
+    private fun getUserDoc(): DocumentReference? =
+        getUser()?.let { user ->
+            firestore
+                .collection(UserDataFirebase.ROOT)
+                .document(user.uid)
+        }
+
+    public fun uploadProfile(profile : ImmutableProfile) {
```
arnauds5: It could be usefull to have some feedback about the success/failure for both upload/get.
For example on others databse we used some callback :
```
uploadProfile(
    profile : ImmutableProfile,
    onSuccess: () -> Unit,
    onFailure: (Exception) -> Unit,
) { ... }
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+        val profile = hashMapOf(
+            "firstName" to profile.firstName,
+            "lastName" to profile.lastName,
+            "age" to profile.age,
+            "gender" to profile.gender.toString(),
+            "phoneNumber" to profile.phoneNumber,
+            "role" to profile.role.toString(),
+            "sciper" to profile.sciper
+        )
+        firestore.collection(ROOT).document(getUser()?.uid!!)
```
arnauds5: Even if the user should be signed in now, throwing a NPE (with !!) could be dangerous.
Maybe giving an Exception to a onFailure callback could be better.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+                    Log.d(TAG, "DocumentSnapshot data: ${document.data}")
+                    val firstName = document.data?.get("firstName")
+                    val lastName = document.data?.get("lastName")
+                    val ageLong = document.data?.get("age")
+                    val genderString = document.data?.get("gender")
+                    val gender = if(genderString?.equals(Gender.MALE.toString())!!) Gender.MALE else if(genderString?.equals(Gender.FEMALE.toString())!!)  Gender.FEMALE else Gender.OTHER
+                    val phoneNumber = document.data?.get("phoneNumber")
+                    val roleString = document.data?.get("role")
+                    val sciperLong = document.data?.get("sciper")
+                    val role = if (roleString == "Teacher") Role.TEACHER else if (roleString == "Student") Role.STUDENT else Role.OTHER
```
arnauds5: I had the same problem of converting raw String to enum for Candidature.State and use this :
https://github.com/asfoury/projmag/blob/a057278f6c3808c8efeec8d1454b11d70c6bdeaf/app/src/main/java/com/sdp13epfl2021/projmag/model/Candidature.kt

Maybe this would be more cleaner (and moving it to Role will be more readable),
(if upload is changed with .name instead of toString())
```
val role = when (roleString) {
    Role.TEACHER.name -> Role.TEACHER
    Role.STUDENT.name -> Role.STUDENT
    Role.OTHER.name -> Role.OTHER
    else -> null //the field could be not valid ? -> return null
}
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+    public fun uploadProfile(profile : ImmutableProfile) {
+        val profile = hashMapOf(
+            "firstName" to profile.firstName,
+            "lastName" to profile.lastName,
+            "age" to profile.age,
+            "gender" to profile.gender.toString(),
+            "phoneNumber" to profile.phoneNumber,
+            "role" to profile.role.toString(),
```
arnauds5: Maybe using this could be better (need to change on the get if changed). (same for gender)
```
            "role" to profile.role.name,
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+                    val roleString = document.data?.get("role")
+                    val sciperLong = document.data?.get("sciper")
+                    val role = if (roleString == "Teacher") Role.TEACHER else if (roleString == "Student") Role.STUDENT else Role.OTHER
+                    val age = Integer.valueOf(ageLong.toString())
+                    var sciper : Int = 0
+                    try {
+                        sciper = Integer.valueOf(sciperLong.toString())
+                    } catch(exep : NumberFormatException) {
+                        sciper = 123456
+                    }
```
arnauds5: sciper could be null, maybe its better to set it to null than putting default value

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+    public fun getProfile(onSuccess: (profile : ImmutableProfile?) -> Unit) {
+        val docRef =  firestore.collection(ROOT).document(getUser()?.uid!!)
+        docRef.get()
+            .addOnSuccessListener { document ->
+                if (document != null) {
+                    Log.d(TAG, "DocumentSnapshot data: ${document.data}")
+                    val firstName = document.data?.get("firstName")
+                    val lastName = document.data?.get("lastName")
+                    val ageLong = document.data?.get("age")
+                    val genderString = document.data?.get("gender")
```
arnauds5: We can access fields directly with this :
`val firstName: String? = document["lastName"] as? String`

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,110 @@
...
+                    val age = Integer.valueOf(ageLong.toString())
+                    var sciper : Int = 0
+                    try {
+                        sciper = Integer.valueOf(sciperLong.toString())
+                    } catch(exep : NumberFormatException) {
+                        sciper = 123456
+                    }
+
+
+                    val resProfile = ImmutableProfile.build(lastName as String, firstName as String, age, gender, sciper, phoneNumber as String, role)
```
arnauds5: Unsafe cast could throw an exception, a try catch could avoid crashing the app if one field coulnd't be cast.
Or checking variable before, for example
```
if (firstName is String && ...) {
    val resProfile = ...
} else {
    ...
}
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: Good suggestion will add this to the method

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: Can it be dangerous?

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: I agree this code looks cleaner will update the code

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
arnauds5: It shouldn't happen in normal cases, but it is safer to avoid throwing NPE.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: ImmutableProfile uses Int and not Int? as the type of the sciper so I cant set it to null

arnauds5: https://github.com/asfoury/projmag/blob/main/app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProfile.kt
In main, sciper is an int?

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: 👍

app/src/main/res/layout/activity_profile_page.xml
```diff
@@ -162,17 +162,12 @@
 
         android:background="@color/white"
         android:orientation="horizontal">
-        <EditText
-            android:id="@+id/info_profile_description"
```
arnauds5, line 169: Because of this field removed, the ProfilePageActivitityTest failed, the part of the test that used the description can be removed.

asfoury: I will remove it!

asfoury:
I have worked a lot of hours on trying to improve the code coverage but I was unable to achieve that, I have a test that's passing locally but wont pass on Cirrus and I tried multiple fixes but they all did not work.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,142 @@
...
+            "lastName" to profile.lastName,
+            "age" to profile.age,
+            "gender" to profile.gender.name,
+            "phoneNumber" to profile.phoneNumber,
+            "role" to profile.role.name,
+            "sciper" to profile.sciper
+        )
+        val id = getUser()?.uid
+        if (id != null) {
+            firestore.collection(ROOT).document(id)
```
arnauds5, line 49: You could reuse the getUserDoc() function (or delete the getUserDoc if it is not used).

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,142 @@
...
+                }
+        } else {
+            Log.d(TAG, "Unable to get the uid from firebase")
+        }
+    }
+
+    public fun getProfile(onSuccess: (profile: ImmutableProfile?) -> Unit) {
+        val userUid = getUser()?.uid
+        if (userUid != null) {
+            val docRef = firestore.collection(ROOT).document(userUid)
```
arnauds5, line 65: You could reuse the getUserDoc() function (or delete the getUserDoc if it is not used).

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,142 @@
...
+                            Integer.valueOf(sciperLong.toString())
+                        } catch (exep: NumberFormatException) {
+                            0
+                        }
+
+
+                        when (val resProfile = ImmutableProfile.build(
+                            lastName ?: "null",
+                            firstName ?: "null",
+                            age,
```
arnauds5, line 105: Instead of returning an incomplete profile when some fields are not valid, it will maybe make more sense to return a null profile to know when a profile is not valid.
Even if "null" name or 0 age look invalid for us, we can't know at 100% if it's a real name or if there was an error, in our code.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,142 @@
...
+                            Role.STUDENT.name -> Role.STUDENT
+                            else -> Role.OTHER
+                        }
+
+
+                        var age: Int = try {
+                            Integer.valueOf(ageLong.toString())
+                        } catch (excep: NumberFormatException) {
+                            0
+                        }
```
arnauds5, line 93: This kind of conversion seems a little weird and we can't know if the field was invalid after this point.
If we return null when a profile is invalid, we could use use this and do null check ?
`val age: Int? = (document["age"] as? Long)?.toInt()`

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
```diff
@@ -0,0 +1,142 @@
...
+                                Log.d(TAG, "Failure reason : ${resProfile.reason}")
+                            }
+                        }
+
+                    } else {
+                        Log.d(TAG, "No such document")
+                    }
+                }
+                .addOnFailureListener { exception ->
+                    Log.d(TAG, "get failed with ", exception)
```
arnauds5, line 124: Instead of just using logging, adding a new argument with a failure callback could be great. (Such as in other database function)

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
```diff
@@ -30,23 +43,73 @@ class ProfilePageActivity : AppCompatActivity() {
...
             val intent = Intent(this,ProjectsListActivity::class.java)
             startActivity(intent)
             finish()
         }
+
+    }
+
+    private fun loadUserProfile(profile : ImmutableProfile?) {
+        if (profile != null) {
+            Log.d(ContentValues.TAG, "NOT NULL")
```
arnauds5, line 73: Using ContentValues.TAG as tag name for logging is strange.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
```diff
@@ -30,23 +43,73 @@ class ProfilePageActivity : AppCompatActivity() {
...
+            findViewById<EditText>(R.id.profile_firstname).setText(profile.firstName)
+            findViewById<EditText>(R.id.profile_lastname).setText(profile.lastName)
+            findViewById<EditText>(R.id.profile_age).setText(profile.age.toString())
+            findViewById<EditText>(R.id.profile_genre).setText(profile.gender.toString())
+            findViewById<EditText>(R.id.profile_phone_number).setText(profile.phoneNumber)
+            findViewById<EditText>(R.id.profile_sciper).setText(profile.sciper.toString())
+        } else {
+            Log.d(ContentValues.TAG, "NULL")
```
arnauds5, line 81: Same as previous comment.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
```diff
@@ -30,23 +43,73 @@ class ProfilePageActivity : AppCompatActivity() {
...
     }
+
+    private fun createProfileFromFields() : ImmutableProfile? {
+        val firstName = findViewById<EditText>(R.id.profile_firstname).text.toString()
+        val lastName = findViewById<EditText>(R.id.profile_lastname).text.toString()
+        val age = findViewById<EditText>(R.id.profile_age).text.toString()
+        val gender = if (findViewById<EditText>(R.id.profile_genre).text.toString() == Gender.MALE.toString())  Gender.MALE
+        else {
+            if (findViewById<EditText>(R.id.profile_genre).text.toString() == Gender.FEMALE.toString()) Gender.FEMALE else Gender.OTHER
+        }
```
arnauds5, line 101: There is some code dupplication, maybe using a when (...) {...} would be more readable.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
```diff
@@ -18,6 +28,9 @@ class ProfilePageActivity : AppCompatActivity() {
     lateinit var buttonSubChange: Button
     private val pickImage = 0
     private var imageUri: Uri? = null
+
+    private var userProfileUploader : UserProfileDatabase? = null
```
PaulCoral: seems unused

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
```diff
@@ -30,23 +43,73 @@ class ProfilePageActivity : AppCompatActivity() {
...
         buttonAddCv.setOnClickListener{
             val intent = Intent(this,CVCreationActivity::class.java)
             startActivity(intent)
             finish()
         }
         buttonSubChange.setOnClickListener{
+            val profile = createProfileFromFields()
+            if(profile != null) {
+                UserProfileDatabase(Firebase.firestore, Firebase.auth).uploadProfile(profile, {}, {})
+            }
```
PaulCoral, line 63: Could be rewrite as
```kotlin
createProfileFromFields()?.{ profile -> UserProfileDatabase(..) }
```
However it is subjective, it just remove one line of code

PaulCoral (approved):
Seems okay, except a few **minor issues** (see comments)
Works on emulator

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: I agree, will fix the problem

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: I agree, will fix the problem

app/src/main/java/com/sdp13epfl2021/projmag/database/UserProfileDatabase.kt
asfoury: I agree I can do that next sprint because I wont have the time to modify the code and the tests before tomorrows sprint

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
asfoury: I agree I can do that next sprint because I wont have the time to modify the code and the tests before tomorrows sprint

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
asfoury: I agree I can do that next sprint because I wont have the time to modify the code and the tests before tomorrows sprint

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
asfoury: Yes its unused I will remove it

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProfilePageActivity.kt
asfoury: I tried to write it like that but it did not work maybe I am doing something wrong?
Will keep it like that in the mean time.

arnauds5 (approved):
Seems to work, even if some modifications need to be done for next weeks.

PR #141, opened 2021-04-23
Add listeners to correctly loads projects in total offline mode

### LINKED ISSUE #139: Projects lists not loading projects if offline (and firebase cache disabled)
If we disable firebase cache and disable internet connection.
The ProjectListActivity won't load the projects from the offline database.
The offline database is correctly loading projects but only a few millisecondes after the first call.

arnauds5:
Add listeners to correctly loads projects in total offline mode.


kaourintin (approved):
The bug seems to be fixed on  my machine.

app/src/main/java/com/sdp13epfl2021/projmag/adapter/ProjectAdapter.kt
```diff
@@ -1,6 +1,6 @@
 package com.sdp13epfl2021.projmag.adapter
 
-import android.content.Context
```
PaulCoral, line 3: Won't compile as Context is not imported, and it is used in the constructor

PaulCoral (changes requested):
Won't compile, see comment

PaulCoral (approved):
Hard to test well on my emulator, but it worked without issue
Fine for me

PR #142, opened 2021-04-24
QR code generation working

asfoury:


app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -327,19 +336,27 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+        else if(item.itemId == R.id.generateQRCodeButton) {
+            val linkToSend = getDynamicLink()
+
+            val qrImage = QRCode.from(linkToSend.toString()).withSize(800,800)
+            val stream = ByteArrayOutputStream()
+            qrImage.bitmap().compress(Bitmap.CompressFormat.PNG, 100, stream)
+            val byteArray: ByteArray = stream.toByteArray()
+
+            val intent = Intent(this, QRCodeActivity::class.java)
+            intent.putExtra("qrcode",byteArray)
```
arnauds5, line 418: Even if the byte array is small (~8KB), maybe moving the qr generation to QRCodeActivity and only sending the url in the the intent, could be better.

app/src/main/java/com/sdp13epfl2021/projmag/activities/QRCodeActivity.kt
```diff
@@ -0,0 +1,20 @@
...
+
+
+class QRCodeActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_q_r_code)
+
+        val imageView = findViewById<ImageView>(R.id.qrcode_image)
+        val byteArray = intent.getByteArrayExtra("qrcode")
+        val bmp = BitmapFactory.decodeByteArray(byteArray, 0, byteArray!!.size)
```
PaulCoral: the `byteArray!!.size` throw a NullPointerException which cause the app to crash in case of issue.
Maybe consider using :
```kotlin
byteArray?.let{ bytes ->
    BitmapFactory....
    imageView....
}
```
Or maybe just `finish()` with a `Toast` message if the `byteArray == null`

PaulCoral (changes requested):
unwanted NullPointerException can arise, and can make the app crash

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -319,6 +319,15 @@ class ProjectInformationActivity : AppCompatActivity() {
         }
     }
 
+    private fun getDynamicLink() : Uri {
```
lpatel99: I would just suggest changing the name of this function to createDynamicLink() for clarity but this is unimportant.

lpatel99 (commented):
Nothing else to say that has not already been said.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
asfoury: I agree this could be better but this would be a optimization that would not give any benefit because as you have said the byte array is small we can add that to the improvements column on the project board

app/src/main/java/com/sdp13epfl2021/projmag/activities/QRCodeActivity.kt
asfoury: I agree, and will add this improvement

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
asfoury: I agree the name suggested is more clear!

PaulCoral (approved):
Good for me, after the last changes

lpatel99 (approved):

arnauds5 (approved):

PR #144, opened 2021-04-27
added ability to apply/unapply

lpatel99:


app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataDatabase.kt
```diff
@@ -81,4 +80,25 @@ interface UserDataDatabase {
...
+     * @param cv the cv to push to the database
+     * @param onSuccess called on success
+     * @param onFailure called with an exception on failure
+     */
+    fun applyUnapply(
+        applied: Boolean,
+        project: ProjectId,
+        onSuccess: () -> Unit,
+        onFailure: (Exception) -> Unit
+    )
```
arnauds5: The documentation do not match with the function arguments.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataFirebase.kt
```diff
@@ -129,4 +137,43 @@ class UserDataFirebase(
                 .addOnFailureListener(onFailure)
         }
     }
+
+    override fun applyUnapply(
```
arnauds5, line 142: When a student apply to a project, it should update two firebase document
- the list of applied projects for this particular user
- the list of student that applied to that particular project (CandidatureDatabase)

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +85,39 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+
+    private fun setUpApplyButton(applyButton: Button) {
+        val projectId = projectVar.id
+        val userDataDatabase = Utils.getInstance(this).userDataDatabase
+        var applied = false
+        userDataDatabase.getListOfAppliedToProjects({
+                projectIds ->
+            if (projectIds.contains(projectId))
+                applied = true
+            setApplyButtonText(applyButton, applied)
```
arnauds5: One line if without {} can be error-prone, and the indentation seems a little weird for projectIds.
```suggestion
        userDataDatabase.getListOfAppliedToProjects({ projectIds ->
            applied = projectIds.contains(projectId)
            setApplyButtonText(applyButton, applied)
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataDatabase.kt
```diff
@@ -81,4 +80,25 @@ interface UserDataDatabase {
...
+     * Call `onSuccess` if the operation succeeded
+     * Call `onFailure` with an Exception in case of failure
+     *
+     * @param cv the cv to push to the database
+     * @param onSuccess called on success
+     * @param onFailure called with an exception on failure
+     */
+    fun applyUnapply(
+        applied: Boolean,
+        project: ProjectId,
```
arnauds5:
```suggestion
        projectId: ProjectId,
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataFirebase.kt
```diff
@@ -129,4 +137,43 @@ class UserDataFirebase(
...
+        unapply: Boolean,
```
arnauds5: This argument name does not match the interface and maybe is the opposite value ? (applied == !unapply)

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataFirebase.kt
```diff
@@ -129,4 +137,43 @@ class UserDataFirebase(
...
+                        temp.add(projectId)
+                    val newList: List<ProjectId> = temp.toList()
+                    doc.set(
+                        hashMapOf(
+                            "applied to" to newList
+                        ), SetOptions.merge()
+                    ).addOnSuccessListener { onSuccess() }.addOnFailureListener(onFailure)
+                },
+                onFailure
+            )
```
arnauds5: This could be simplified to an atomic update (and avoid nested database calls), for example by this
```suggestion
            val fieldValue = if (unapply) {
                FieldValue.arrayRemove(projectId)
            } else {
                FieldValue.arrayUnion(projectId)
            }
            doc.update(APPLIED_TO_FIELD, fieldValue)
                .addOnSuccessListener { onSuccess() }
                .addOnFailureListener(onFailure)
```

arnauds5:
The tests build failed because FakeUserDataDatabase need to have all the function of the UserDataDatabse interface..

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +85,37 @@ class ProjectInformationActivity : AppCompatActivity() {
         }
     }
 
+    private fun setApplyButtonText(applyButton: Button, applied: Boolean) {
+        if (applied)
+            applyButton.text = "UNAPPLY"
+        else
+            applyButton.text = "APPLY"
```
arnauds5: Using resources strings is better in general instead of raw literal.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataFirebase.kt
lpatel99: I agree, however considering this is not my task for this week and I have already spent a lot of time working on this and have not finished my tasks, I would rather deal with that in a task next week.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataDatabase.kt
```diff
@@ -81,4 +80,33 @@ interface UserDataDatabase {
...
+     * Apply to the given project by pushing it to
+     * Call `onSuccess` if the operation succeeded
+     * Call `onFailure` with an Exception in case of failure
+     *
+     * @param applied function applies if true, else unapplies
+     * @param onSuccess called on success
+     * @param onFailure called with an exception on failure
+     */
+    fun applyUnapply(
+        applied: Boolean,
```
arnauds5, line 94: So, if applied is true it should add the projectId,
and if applied is false it will remove it ?

arnauds5: So, if applied is true it should add the projectId,
and if applied is false it will remove it ?

lpatel99: I changed the variable to apply in the implementation of the interface but forgot to change it here

arnauds5: Now it's the documentation that is not the same as the function.
And I'm still not sure that I understood what this function should do when it's true/false.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +91,40 @@ class ProjectInformationActivity : AppCompatActivity() {
         }
     }
 
+    private fun setApplyButtonText(applyButton: Button, applied: Boolean?) {
+        if (applied == null)
+            applyButton.text = LOADING_STRING
+        else if (applied)
+            applyButton.text = UNAPPLY_STRING
+        else
+            applyButton.text = APPLY_STRING
```
arnauds5: It seems that there is a little code dupplicate (applyButton.text = ...), maybe this will be shorter : 
```suggestion
        applyButton.text = when (applied) {
            null -> LOADING_STRING
            true -> UNAPPLY_STRING
            false -> APPLY_STRING
        }
```

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataDatabase.kt
lpatel99: Indeed you are right, I fixed it.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +91,39 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+        setApplyButtonText(applyButton,null)
+        userDataDatabase.getListOfAppliedToProjects({ projectIds ->
+            alreadyApplied = projectIds.contains(projectId)
+            setApplyButtonText(applyButton, alreadyApplied)
+        },{})
+
+        applyButton.setOnClickListener {
+            userDataDatabase.applyUnapply(
+                !alreadyApplied,
+                projectVar.id,
```
arnauds5:
```suggestion
               projectId,
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +91,39 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+                {
+                    Toast.makeText(this, getString(R.string.success), Toast.LENGTH_SHORT).show()
```
arnauds5: There is a function showToast(msg) that you could use here.
```suggestion
                    showToast(getString(R.string.success))
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -88,6 +91,39 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+                    alreadyApplied = !alreadyApplied
+                    setApplyButtonText(applyButton, alreadyApplied)
+                },
+                {Toast.makeText(this, getString(R.string.failure), Toast.LENGTH_SHORT).show()}
```
arnauds5: Same  as previous comment.
```suggestion
                {showToast(getString(R.string.failure))}
```

asfoury (commented):
When running the app on the emulator and testing the apply button I get a toast with the message failed.
Also I tried applying to the greyed out projects in the bottom of the projects list and was able to apply.

asfoury (approved):
I tested the code now and it works, and now the apply button is greyed out when the project is taken.

app/src/main/java/com/sdp13epfl2021/projmag/database/UserDataFirebase.kt
```diff
@@ -129,4 +138,35 @@ class UserDataFirebase(
...
+    ) {
+        getUserDoc()?.let { doc ->
+            val fieldValue = if (apply) {
+                FieldValue.arrayUnion(projectId)
+            } else {
+                FieldValue.arrayRemove(projectId)
+            }
+            doc.update(APPLIED_TO_FIELD, fieldValue)
+                .addOnSuccessListener { onSuccess() }
+                .addOnFailureListener(onFailure)
```
arnauds5, line 156: If the /user-data/{userID} document does not exists, the applied list is not updated.
You can either try to create it if it does not exists, or we can also rely on the fact that the document will be created after the user register.

arnauds5 (approved):
Looks good to me.
The problem of the applied list not updated can be either fix by creating the user document when the user register the first time, or by checking if it exists.

PR #145, opened 2021-04-28
split interface for Student and professor

giogio21:


app/src/main/java/com/sdp13epfl2021/projmag/UserTypeChoice.kt
```diff
@@ -0,0 +1,32 @@
...
+import androidx.appcompat.app.AppCompatActivity
+import com.sdp13epfl2021.projmag.activities.ProfilePageActivity
+
+class UserTypeChoice : AppCompatActivity() {
+    lateinit var ProfButton: RadioButton
+    lateinit var StuButton: RadioButton
+
+    companion object{
+        var data: Boolean = false
+    }
```
arnauds5: This meaning of this variable is not clear at all.

arnauds5 (changes requested):

app/src/main/java/com/sdp13epfl2021/projmag/UserTypeChoice.kt
giogio21: I renamed the variable

app/src/main/java/com/sdp13epfl2021/projmag/UserTypeChoice.kt
```diff
@@ -0,0 +1,32 @@
...
+import android.content.Intent
+import android.os.Bundle
+import android.util.Log
+import android.widget.RadioButton
+import androidx.appcompat.app.AppCompatActivity
+import com.sdp13epfl2021.projmag.activities.ProfilePageActivity
+
+class UserTypeChoice : AppCompatActivity() {
+    lateinit var ProfButton: RadioButton
+    lateinit var StuButton: RadioButton
```
arnauds5: If theses buttons are only used in onCreate, we could remove theses var and use val profButton = ... in onCreate.
Variable name should start with lowercase for camelCase. (also, Stu is maybe a too short abreviation).

app/src/main/java/com/sdp13epfl2021/projmag/UserTypeChoice.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    companion object{
+        var isProfessor: Boolean = false
+    }
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_user_type2)
```
arnauds5: activity_user_type2.xml
Is there a reason for having a "2" in this layout name ?

giogio21: No, it is a typo

app/src/main/res/layout/activity_user_type_choice.xml
```diff
@@ -0,0 +1,50 @@
...
+            android:layout_height="wrap_content"
+            android:layout_alignParentStart="true"
+            android:layout_alignParentTop="true"
+            android:layout_marginStart="131dp"
+            android:layout_marginTop="265dp"
+            android:text="@string/selection"
+            android:textSize="20sp" />
+
+
+    </RelativeLayout>
```
arnauds5: The UI looks fine but the code is a little "aproximatif" with absolute positions/etc, it could be better to use the constraint layout such as in the suggestion.
Because if we use different screen size, the elements will not be centered.

image

(If you use this suggestion, the tests would need to be modified because of the relative layout removed)
```suggestion
    <RadioGroup
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:layout_alignParentStart="true"
        android:layout_alignParentTop="true"
        android:layout_marginTop="20dp"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintHorizontal_bias="0.5"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toBottomOf="@+id/textViewUserType">

        <RadioButton
            android:id="@+id/radioProfessorType"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:text="@string/professor" />

        <RadioButton
            android:id="@+id/radioStudentType"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:text="@string/student" />
    </RadioGroup>

    <TextView
        android:id="@+id/textViewUserType"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:text="@string/selection"
        android:textSize="20sp"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintHorizontal_bias="0.5"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:layout_constraintVertical_bias="0.4" />
```

app/src/androidTest/java/com/sdp13epfl2021/projmag/UserTypeChoiceTest.kt
```diff
@@ -0,0 +1,97 @@
...
+                    childAtPosition(
+                        withClassName(`is`("android.widget.RelativeLayout")),
+                        0
+                    ),
+                    1
+                ),
+                isDisplayed()
+            )
+        )
+        materialRadioButton2.perform(click())
```
arnauds5, line 83: Same as the previous comment : 
assertFalse(UserTypeChoice.isProfessor)

arnauds5 (commented):
UserTypeChoice.kt could be move to the activities directory.
activity_user_type_choice.xml has some absolute positions that won't be centered on all screens.

app/src/androidTest/java/com/sdp13epfl2021/projmag/UserTypeChoiceTest.kt
```diff
@@ -0,0 +1,97 @@
...
+                        withClassName(`is`("android.widget.RelativeLayout")),
+                        0
+                    ),
+                    0
+                ),
+                isDisplayed()
+            )
+        )
+        materialRadioButton.perform(click())
+
```
arnauds5: This only tests that the button can be clicked.
It would be better to also check that it does modify the isProfessor value.
```
assertTrue(UserTypeChoice.isProfessor)
```

arnauds5 (approved):
For the moment the isProfessor variable is only set when the user first register.
In an another task for next weeks, we need to save/load this.

app/src/main/java/com/sdp13epfl2021/projmag/activities/UserTypeChoice.kt
```diff
@@ -0,0 +1,30 @@
+package com.sdp13epfl2021.projmag.activities
+
+import android.content.Intent
+import android.os.Bundle
+import android.widget.RadioButton
+import androidx.appcompat.app.AppCompatActivity
+import com.sdp13epfl2021.projmag.R
+
+class UserTypeChoice : AppCompatActivity() {
```
PaulCoral, line 9: It would be better if class name end with `Activity`.

PaulCoral (approved):
It is ok, just check my comment please

PR #169, opened 2021-05-03
add PHD role

giogio21:


lpatel99 (commented):
Considering the issues we've had with espresso test recorder, I would recommend manually writing the tests here, especially considering they will be very simple.

lpatel99 (approved):
All good!

arnauds5 (commented):
With this code, there is no difference between professor and PHD.
Is it planned to give different rights/etc to them or will they stay identical ?

giogio21:
> With this code, there is no difference between professor and PHD.
> Is it planned to give different rights/etc to them or will they stay identical ?

Yes for the moment there is no difference but it plans to give PHD diffrents rights

app/src/androidTest/java/com/sdp13epfl2021/projmag/UserTypeChoiceTest.kt
```diff
@@ -21,84 +21,29 @@ import org.junit.Rule
...
-                    0
-                ),
-                isDisplayed()
-            )
-        )
-        materialRadioButton.perform(click())
+    fun userProf() {
+        onView(withId(R.id.radioProfessorType)).perform(click())
         assertTrue(UserTypeChoice.isProfessor)
+    }
```
PaulCoral, line 35: these kind of tests using singletons are critcized in the code review should be changed, now or once we introduced dependency injections

PaulCoral (approved):
Seems ok, except my previous comment

arnauds5 (approved):

PR #172, opened 2021-05-03
added filter by applied projects

### LINKED ISSUE #160: Add applied to projects to list filter.

lpatel99:


app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
```diff
@@ -22,12 +22,20 @@ import com.sdp13epfl2021.projmag.database.Utils
 import com.sdp13epfl2021.projmag.model.ProjectFilter
 
 class ProjectsListActivity : AppCompatActivity() {
+
     private lateinit var itemAdapter: ProjectAdapter
     private lateinit var recyclerView: RecyclerView
-
+    private lateinit var appliedProjects: MutableList<ProjectId>
```
arnauds5: The creation of this variable could be done directly here and with val, instead of of a var in onCreate.
```
    private val appliedProjects: MutableList<ProjectId> = ArrayList()
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
```diff
@@ -22,12 +22,20 @@ import com.sdp13epfl2021.projmag.database.Utils
...
+    private lateinit var utils: Utils
```
arnauds5, line 29: If this variable is only used for ProjectAdapter in onCreate, there is no reason to store it in a var ?

arnauds5 (approved):
Looks good to me.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
lpatel99: good point! I've changed that

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
lpatel99: Are you talking utils?
If so, it is also used in updateAppliedProjects() which is called not only in onCreate but also in onResume()

arnauds5: yes, I didn't see it the first time

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
```diff
@@ -22,12 +22,20 @@ import com.sdp13epfl2021.projmag.database.Utils
...
+
     private lateinit var itemAdapter: ProjectAdapter
     private lateinit var recyclerView: RecyclerView
-
+    private val appliedProjects: MutableList<ProjectId> = ArrayList()
+    private lateinit var utils: Utils
+
+    private fun updateAppliedProjects() {
+        utils.userDataDatabase.getListOfAppliedToProjects({ list ->
+            appliedProjects.removeAll { true }
```
PaulCoral: Wouldn't it be better ?
```kotlin
mutableList.clear()
```

PaulCoral (changes requested):
Everything looks good, just a small change

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectsListActivity.kt
lpatel99: ah yes definitely! I could not remember the function

PaulCoral (approved):
Fine after last changes

kaourintin (approved):
All good!

PR #173, opened 2021-05-04
personal list of projet

giogio21:
As professor when you submit a project , a copy is add to your personal project list. You can access it by clicking on the icon list in the menu bar


arnauds5 (commented):
Instead of creating a new UI/etc for listing our own projects, a filter could be used with the existing ProjectsListActivity. This will reduce code duplications and it will be more simple.

giogio21:
> Instead of creating a new UI/etc for listing our own projects, a filter could be used with the existing ProjectsListActivity. This will reduce code duplications and it will be more simple.

Yes I think it is better , I will do this that way

PR #176, opened 2021-05-05
Preferences

PaulCoral:
Fix from code review


app/src/main/java/com/sdp13epfl2021/projmag/activities/PreferencesActivity.kt
```diff
@@ -0,0 +1,118 @@
...
+     */
+    private lateinit var applied: CheckBox
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.project_filter_settings)
+
+        // Update the view for this activity
+        findViewById<TextView>(R.id.filter_title).text = getString(R.string.preferences)
+        findViewById<View>(R.id.filter_preferences_switch).visibility = View.GONE
```
asfoury, line 46: When I test on the emulator I think that the expected behavior is inverted when I activate preferences the preference options disappear and when I deactivate it they appear

asfoury (approved):

arnauds5 (approved):
When we enable preference, then click on a projet then return, the preference is not enabled and we need to reclick.
Except this, everythings looks fine.

PR #178, opened 2021-05-06
applying adds dummy candidature but unapplying does not yet remove

lpatel99:


app/src/main/java/com/sdp13epfl2021/projmag/database/FirebaseCandidatureDatabase.kt
```diff
@@ -121,4 +136,21 @@ class FirebaseCandidatureDatabase(
...
+    ) {
+        getDoc(candidature.projectId)
+            .let { doc ->
+                val fieldValue = FieldValue.arrayRemove(candidature.userID)
+                doc.update(APPLIED_TO_FIELD, fieldValue)
+                    .addOnSuccessListener { onSuccess() }
+                    .addOnFailureListener(onFailure)
+            }
+
+    }
```
arnauds5, line 152: Because the candidature is a map of (UserID to State), we need to use delete instead of arrayRemove.
```
override fun removeCandidature(
        projectID: ProjectId,
        userID: String,
        onSuccess: () -> Unit,
    ) {
        getDoc(projectID)
            ?.update(userID, FieldValue.delete())
            ?.addOnSuccessListener { onSuccess() }
            ?.addOnFailureListener(onFailure)
    }
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -108,26 +112,45 @@ class ProjectInformationActivity : AppCompatActivity() {
...
         applyButton.setOnClickListener {
             userDataDatabase.applyUnapply(
                 !alreadyApplied,
-               projectId,
+                projectId,
                 {
                     showToast(getString(R.string.success), Toast.LENGTH_SHORT)
                     alreadyApplied = !alreadyApplied
                     setApplyButtonText(applyButton, alreadyApplied)
+                    if (!alreadyApplied) {
```
arnauds5, line 138: Because the alreadyApplied was already inverse, the if should not be inversed.
```suggestion
                    if (alreadyApplied) {
```

PR #191, opened 2021-05-07
Add KDoc + rename classes + reorganization of the Database package

arnauds5:
**This PR does not only contain KDoc changes.**
Some classes were also renamed and the package database was reorganized into multiple levels.

The database package is now organized into : 
```
database/
    interfaces/
        CandidatureDatabase
        FileDatabase
        MetadataDatabase
        ProjectDatabase
        UserdataDatabase
    impl/
        firebase/
            FirebaseCandidatureDatabase
            FirebaseFileDatabase
            FirebaseMetadataDatabase
            FirebaseProjectDatabase
            FirebaseUserdataDatabase
            UserProfileDatabase
        cache/
            CacheProjectDatabase
            OfflineCachedCandidatureDatabase
            OfflineCachedUserdataDatabase
            OfflineProjectDatabase
    LocalFileUtils
    ProjectChange
    ProjectUploader
    Utils
```

Things that are need to be change in another PR : 
- Add an abstraction to UserProfileDatabase and rename it (with KDoc)
- Merge userdata & userProfile ?
- Remote Utils and use dependency injection instead
- Save preferences to local storage for offline use
- Abstract how the local userID is get (after dependency injection)


.idea/codeStyles/Project.xml
```diff
@@ -3,9 +3,18 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="java.util" withSubpackages="false" static="false" />
-          <package name="kotlinx.android.synthetic" withSubpackages="true" static="false" />
-          <package name="io.ktor" withSubpackages="true" static="false" />
+          <package name="java.util" alias="false" withSubpackages="false" />
```
PaulCoral: I don't think this file is needed

There's also another `.idea/*` file

app/src/main/java/com/sdp13epfl2021/projmag/database/LocalFileUtils.kt
```diff
@@ -3,47 +3,46 @@ package com.sdp13epfl2021.projmag.database
...
-                it.writeObject(data)
-                return true
-            }
-        } catch (e: Exception) {
-            e.printStackTrace()
-            return false
+/**
+ * This serializable class can be used to wrap a list of String.
+ */
+data class SerializedStringListWrapper(val list: List<String>) : Serializable
```
PaulCoral, line 10: why is this class added ?
I can't find its previous declaration

PaulCoral (approved):
Everything is fine, just look at my few comments, these are minor issues. I let you judge if you have to apply these changes.

app/src/main/java/com/sdp13epfl2021/projmag/database/LocalFileUtils.kt
arnauds5: It was in OfflineCachedUserDataDatabase.
A List<*> is not serializable directly, so I use this class to wrap a list of string. 
Because it could be used in other offline databases/etc, I move it to LocalFileUtils.

.idea/codeStyles/Project.xml
arnauds5: I reverted the change to this file.
Should we delete the .idea directory and add it to .ignore ?
If yes, it will be better to do it in a more global PR, for example for cleaning the whole project.

lpatel99 (approved):
All good. If you haven't already done it, before merging please reformat the code and optimize imports. You can do this by right clicking on the "java" folder.

PR #193, opened 2021-05-09
Fix activity_project_creation.xml UI issues

### LINKED ISSUE #192: The video feedback on the project creation is not clearly visible because of overlapping buttons.
image

PaulCoral:
FIXED

* the video goes bellow the button
* the buttons were wrongly placed


arnauds5 (approved):

asfoury (approved):
The buttons look a lot better now, maybe in the next sprints we can try to find a way for the video controller buttons to appear in a better position than where they are appearing now

PR #200, opened 2021-05-12
add class allow find orientation

### LINKED ISSUE #185: Find a way to have the phone orientation

giogio21:


kaourintin (commented):
I think it wouldn't be that hard to implement tests that check if the position outputted by this class is the same as the default position when the android studio simulator is launched

lpatel99 (approved):
All good, just put the orientation activity in the appropriate package before merging.

PaulCoral (approved):
For me it is fine for now
We need to go further when we will implement the other part of AR

PR #204, opened 2021-05-13
Add Hilt DI to all activites and tests (at least where it is currently possible)

### LINKED ISSUE #184: Dependency Injection, where possible

PaulCoral:


PaulCoral:
As the issue do not seem to have an easy solution, I close this PR, and start a new branch

arnauds5 (approved):
looks good to me.

~Now that we can inject database directy, maybe we could remove all the database passed in constructor, for example in CandidatureAdapter, ProjectAdapter, etc. But it would take some time for nearly the same, and maybe not worth it.~ (as he answered in discord, that could lead to some problems with singleton/context)

arnauds5 (commented):
I have a question about the currentUserId that I'm not sure what the code do now :
If a new user register for the first time, the injected userId is not updated ?

PaulCoral:
@arnauds5 
> I have a question about the currentUserId that I'm not sure what the code do now :
> If a new user register for the first time, the injected userId is not updated ?

Good question, It depends on what we want. The annotation `@Singleton` in the UserIdModule tells the Hilt framework that the value of the will be the same (not recalled or recomputed) for the lifetime of the application (generated in the `onCreate()` method of the **Application**), so the userId is generated each time we launch the App
So maybe I can remove this annotation, so the default behaviour would be (from Hilt website) :

> By default, all bindings in Dagger are “unscoped”. This means that each time the binding is requested, Dagger will create a new instance of the binding.

**EDIT**:
We discussed and found out that at the first log-in we could have issue (empty userId)

.codeclimate.yml
```diff
@@ -11,6 +11,7 @@ checks:
       threshold: 250 # not number of lines, but mass threshold (see default analysis configuration)
 
 exclude_patterns:
-- "**/test/"
-- "**/androidTest/"
-- "**test"
+  - "**/test/"
+  - "**/androidTest/"
+  - "**test"
+  - "**/di/" # Exclude Hilt modules (Dependency Injections)
```
arnauds5: We shouldn't exclude anything according to the email we just received :
_"The line coverage of your tests must be at least 80% of the total code (without any code exclusion)"_
```suggestion
```

PaulCoral: Ok, I did this because they are already excluded from jacoco

app/src/main/java/com/sdp13epfl2021/projmag/database/di/UserdataDatabaseModule.kt
```diff
@@ -16,10 +16,10 @@ import javax.inject.Singleton
 object UserdataDatabaseModule {
     @Provides
     @Singleton
-    fun bindUserdataDatabase(
+    fun providesUserdataDatabase(
         userDB: FirebaseUserdataDatabase,
-        @Named("userID") uid: String,
-        @Named("userRootDir") dir: File
+        @Named("currentUserId") uid: String,
+        @Named("fileCacheUserDB") dir: File
```
lpatel99, line 22: Maybe avoid string literals (especially for those you use in more than one place)

lpatel99 (approved):
Otherwise all good and works well!

PR #207, opened 2021-05-13
Add missing database implementation + fix bugs

### LINKED ISSUE #213: Add missing implementations of databases

### LINKED ISSUE #216: When creating/modifying the profile, if we are a teacher (sciper disabled), the code crash.
Because the emtpy String "" can't be converted to a Int, the code crash on ProfilePageActivity on the createProfileFromFields for the sciper.

### LINKED ISSUE #217: If the profile contains a null sciper (teacher), the getProfile of FIrebaseUseradataDatabase failed.

arnauds5:
Fix https://github.com/asfoury/projmag/issues/216
Fix https://github.com/asfoury/projmag/issues/217
Add missing database implementations.
Add real profile/cv to candidatures.


app/src/androidTest/java/com/sdp13epfl2021/projmag/database/fake/FakeCandidatureDatabase.kt
```diff
@@ -25,13 +25,15 @@ class FakeCandidatureDatabase(
...
-//        candidaturesState[projectId] = newMap1 + (userId to newState)
-//        candidatures[projectId] = newMap2 + (userId to candidature)
-//        onChanges[projectId]?.let { it.forEach { it(projectId, candidatures[projectId]?.values?.toList() ?: emptyList()) } }
+        val newMap1 = candidaturesState[projectId] ?: emptyMap()
+        val newMap2 = candidatures[projectId] ?: emptyMap()
+        candidaturesState[projectId] = newMap1 + (userId to newState)
+        val oldCand = newMap2[userId]
+        if (oldCand != null) {
+            val newCand = Candidature(projectId, userId, oldCand.profile, oldCand.cv, newState)
+            candidatures[projectId] = newMap2 - userId + (userId to newCand)
```
lpatel99: I think that adding an entry with a preexisting key just updates the entry, so don't need the - userId

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCandidatureDatabase.kt
```diff
@@ -81,45 +78,38 @@ class FirebaseCandidatureDatabase(
...
+            profileDeferred.complete(profile)
+            // If the profile is null, we stop to wait because no candidature will be added
+            if (profile == null) {
+                waiting.set(false)
+            }
+        }, { waiting.set(false) }) // If a problem occurs, we stop to wait because no candidature will be added
+
+        userdataDatabase.getCv(userID, { cv ->
+            cvDeferred.complete(cv)
+            // If the cv is null, we stop to wait because no candidature will be added
```
lpatel99, line 95: so do we consider that there needs to already be a profile and cv for a student to apply for a project (i.e. userId alone is not sufficient)?

lpatel99:
Also could you add the necessary deletion of gradle.xml to this PR so that the main works again please (or do you think it is better to do a separate PR) ?

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCandidatureDatabase.kt
arnauds5: getListOfCandidatures will only return a list of complet candidature.
If a student apply to a project, normally its profile/cv should already exists.

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseUserdataDatabase.kt
```diff
@@ -47,6 +47,8 @@ class FirebaseUserdataDatabase(
         const val PREF_FIELD = "preferences"
 
         const val USER_PROFILE = "user-profile"
+
+        private val AUTH_EXCEPTION: Exception = SecurityException("User need to be authenticated")
```
kaourintin:
```suggestion
        private val AUTH_EXCEPTION: Exception = SecurityException("User needs to be authenticated")
```

app/src/androidTest/java/com/sdp13epfl2021/projmag/database/fake/FakeCandidatureDatabase.kt
arnauds5: I changed that

arnauds5:
> Also could you add the necessary deletion of gradle.xml to this PR so that the main works again please (or do you think it is better to do a separate PR) ?

I removed the idea files and add this directory to the git ignore

kaourintin (approved):

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCandidatureDatabase.kt
arnauds5: It wouldn't make sense to consider candidature with empty profile/cv, if it was the case, the professor won't be able to view who wants to apply, except a random userId.

kaourintin (approved):
Looks good to me, I agree with Arnaud on the fact that when presenting data to the teacher on the application everything should not be allowed to be null

lpatel99 (approved):
Great!

PR #223, opened 2021-05-18
Feature/project storing offline

### LINKED ISSUE #134: Create list of own projects available offline.

### LINKED ISSUE #211: Create a cache manager that cleans up the right cache, manages video size and checks cache availability when a project is added/removed from favourites

kaourintin:
=> memory storing is now handled properly between cacheDir and projDir
=> add a function in database interface for getting the file names
=> we now store our own projects entirely in local permanent storage
=> add testing for favourite button


app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/FileDatabase.kt
```diff
@@ -79,4 +79,7 @@ interface FileDatabase {
         onFailure: (Exception) -> Unit
     )
 
+
+    fun getFileName(fileUrl: String) : String
```
arnauds5: KDoc need to be added.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -417,6 +426,18 @@ class ProjectInformationActivity : AppCompatActivity() {
         }, { showToast(getString(R.string.could_not_download_video), Toast.LENGTH_LONG) })
     }
 
+    private fun deletingAndCopyingVideo(fileUrl: String, deleteDirectory: File, copyDirectory: File){
```
arnauds5: Using names like moveVideoFile(fileUrl: String, fromDirectory: File, toDirectory: File) or other names would be better to understand.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -417,6 +426,18 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+        val file = File(deleteDirectory, fileDB.getFileName(fileUrl))
+        if(file.exists()){
+            val newFile = File(copyDirectory, fileDB.getFileName(fileUrl))
+            file.copyRecursively(newFile, true)
+            file.deleteRecursively()
```
arnauds5: Instead of copying then deleting the file, just moving the file could be cleaner.
For example using Files.move(fromFile, toFile) from com.google.common.io.Files.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -377,23 +380,29 @@ class ProjectInformationActivity : AppCompatActivity() {
...
 
         videosLinks.forEach { link ->
-            if(isFavorite){//storing the video with persistence
+            if(isFavorite || ownProjects.any {project -> project.id == projectVar.id}){//storing the video in the permanent memory
+                deletingAndCopyingVideo(link, cacheDir, projectDir)
                 storingVideo(link, projectDir)
             }else {
+                deletingAndCopyingVideo(link, projectDir, cacheDir)//storing
                 storingVideo(link, cacheDir)
             }
```
arnauds5: This code will try to copy/delete file each time a video is loaded, even if it was already in the correct directory.
Is it the expected behavior ?
And the if-else block seems nearly identical, maybe it could be use as a single block with 2 val for from/to directory ?

arnauds5 (changes requested):

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
kaourintin: the code doesn't move the two files if the file that needs to be moved doesn't exist if(file.exist()) in function move

It can be put with intermediate val from/to but I still need the if else block, I don't feel like creating two extra intermediate vals makes the code better

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
kaourintin: Yes my bad I forgot to push it this way yesterday

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/FileDatabase.kt
```diff
@@ -79,4 +79,12 @@ interface FileDatabase {
         onFailure: (Exception) -> Unit
     )
 
+    /**
+     * Asynchronously get the actual filename from the firebase database
```
arnauds5: It doesn't seems to be asynchronously, is it ?

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/FileDatabase.kt
```diff
@@ -79,4 +79,12 @@ interface FileDatabase {
...
+     *
+     * @param fileUrl the file url
+     * @return the firebase file name
```
arnauds5: The abstracted interface shouldn't mention the implementation (Firebase)

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -417,6 +439,31 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+     *
+     * @param fileUrl url of the file to be moved
+     * @param deleteDirectory directory from which it has to be moved
+     * @param copyDirectory directory in which it is moved to
+     * @return true if move was a success or there was no need to move, false else
+     */
+    private fun movingVideo(fileUrl: String, deleteDirectory: File, copyDirectory: File): Boolean{
+        val file = File(deleteDirectory, fileDB.getFileName(fileUrl))
+        if(file.exists()){
+            val newFile = File(copyDirectory, fileDB.getFileName(fileUrl))
```
arnauds5: Instead of calling fileDB.getFileName(fileUrl) two times, it could be done using an intermediate val for filname,

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -439,6 +454,32 @@ class ProjectInformationActivity : AppCompatActivity() {
...
 
+    /**
+     * moves a video from the delete Directory to the copy Directory
+     *
+     * @param fileUrl url of the file to be moved
+     * @param deleteDirectory directory from which it has to be moved
+     * @param copyDirectory directory in which it is moved to
+     * @return true if move was a success or there was no need to move, false else
+     */
+    private fun movingVideo(fileUrl: String, deleteDirectory: File, copyDirectory: File): Boolean{
```
arnauds5: The return value seems to be unused in the code ?
If it is the case, the function could be simplified by removing the return value.
Or will there be some future use of it ?

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/FileDatabase.kt
```diff
@@ -79,4 +79,12 @@ interface FileDatabase {
         onFailure: (Exception) -> Unit
     )
 
+    /**
+     *  get the actual filename from the database
+     *
+     * @param fileUrl the file url
+     * @return the firebase file name
```
arnauds5: It's still mentioning the firebase implementation.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -394,29 +404,37 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+        //get the favorite list, the applied To list and the own projects list
         userdataDatabase.getListOfFavoriteProjects({ favorites ->
-            addVideoAfterDownloadedWithFavoritePersistence(
-                videosLinks,
-                favorites.contains(projectVar.id)
-            )
-        }, {
+            projectDatabase.getAllProjects({projects ->
+                addVideoDownload(videosLinks, favorites.contains(projectVar.id),
+            projects.filter { project -> project.authorId == FirebaseAuth.getInstance().uid })},
```
arnauds5: If the PR https://github.com/asfoury/projmag/pull/204 is accepted before this one, don't forget to change.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -394,29 +404,37 @@ class ProjectInformationActivity : AppCompatActivity() {
...
 
-    private fun handleVideoWithFavoritePersistenceFiltering(videosLinks: List<String>) {
-        //get the favorite list and when it's available provide it to the function that is
-        //responsible for downloading videos in volatile or non volatile memory
+
+    /**
+     * Function that downloads a video either in the proj dir permanent app memory if it's
+     * a favorite or one of our own projects. Else, the video is stored in the cache dir which is a
+     * memory that is deleted by the phone if it needs it. This function calls a callback and will do
+     * nothing if the userDatabase or the projectDatabase didn't respond for some reason.
```
arnauds5: The documentation about the callback seems a little weird.
What do you mean by "didn't respond for some reason" ?

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -439,6 +457,32 @@ class ProjectInformationActivity : AppCompatActivity() {
         }, { showToast(getString(R.string.could_not_download_video), Toast.LENGTH_LONG) })
     }
 
+    /**
+     * moves a video from the delete Directory to the copy Directory
+     *
+     * @param fileUrl url of the file to be moved
+     * @param deleteDirectory directory from which it has to be moved
+     * @param copyDirectory directory in which it is moved to
+     * @return true if move was a success or there was no need to move, false else
```
arnauds5: The documention was not updated
```suggestion
```

PaulCoral: I didn't edit this fun, so this must be prior this PR

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
kaourintin: I just meant if it returns onFailure() I do nothing in this case

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
kaourintin: Yes we won't be able to miss it anyways when we check the TODO's remaining in the code next week

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
arnauds5: didn't respond and having the failure callback called are not the same.
Also, this function does not call the callback, it's database that calls the callback.

app/src/androidTest/java/com/sdp13epfl2021/projmag/ProjectInformationActivityTest.kt
```diff
@@ -244,6 +246,23 @@ class ProjectInformationActivityTest {
         applyButton.check(matches(withText(applyString)))
     }
 
+
+    @Test
+    //TODO : call the proper string ressource in string.xml w mockito
+    fun clickOnFavoriteButton() {
+        var favoriteString : String = "add project to favorites"
+        var removeFavoriteString : String = "remove project from favorites"
+
```
PaulCoral, line 258: Maybe you could use `ApplicationProvider.getApplicationContext<Context>().getString(....)`

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -439,6 +457,31 @@ class ProjectInformationActivity : AppCompatActivity() {
...
+    private fun movingVideo(fileUrl: String, deleteDirectory: File, copyDirectory: File){
+        val fileName = fileDB.getFileName(fileUrl)
+        val file = File(deleteDirectory, fileName)
+        if(file.exists()){
+            val newFile = File(copyDirectory, fileName)
+            try{
+                Files.move(file, newFile)
+            }catch(e: Exception){
+                
+            }
```
PaulCoral: Catching any `Exception` is usually a bad practice, potentially **hiding a deeper** issue in the code-base.
Please catch an `Exception` as precise as it is possible.
(Also you can show a toast message showing the error if needed)

PaulCoral (changes requested):
Mostly good, just an issue with `try{...}catch(...){...}`

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
kaourintin: The exeption was either launched when the delete and copy directory were identical (so we do nothing because that's fine)
Or if the IO fails somehow (which is not something we can control and I did not think the user needed to be notified of an IO failure)
I added a comment either way in the case of IO and we still do nothing in case of Illegal arg

app/src/androidTest/java/com/sdp13epfl2021/projmag/ProjectInformationActivityTest.kt
kaourintin: Thanks I was actually looking for this

arnauds5 (approved):

PaulCoral (approved):

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
arnauds5: The PR https://github.com/asfoury/projmag/pull/204 is already merged.

PR #242, opened 2021-05-23
Feature/add cv display view

### LINKED ISSUE #235: Add an activity to display the cv

### LINKED ISSUE #238: Firebase getCv is not correctly implemented
java.lang.ClassCastException: java.util.HashMap cannot be cast to com.sdp13epfl2021.projmag.curriculumvitae.PeriodDescription
        at com.sdp13epfl2021.projmag.curriculumvitae.CurriculumVitae.writeToParcel(Unknown Source:43)

### LINKED ISSUE #240: OrientationActivityTest.testDefaultPosition will never pass if our emulator don't have the same "default" position

arnauds5:
The easiest way to show the cv display view, is by checking the candidatures to your own project.
(For the moment you can still apply to your own project, but don't forget to have a profile/cv before)

Fix https://github.com/asfoury/projmag/issues/238.
Fix https://github.com/asfoury/projmag/issues/240.

image
image


app/src/main/java/com/sdp13epfl2021/projmag/adapter/CVAdapter.kt
```diff
@@ -0,0 +1,91 @@
...
+        val line2TextView: TextView = view.findViewById(R.id.cv_display_item_line_2)
+    }
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): CVHolder {
+        val adapterLayout = LayoutInflater.from(parent.context)
+            .inflate(R.layout.cv_display_item, parent, false)
+        return CVHolder(adapterLayout)
+    }
+
+    override fun onBindViewHolder(holder: CVHolder, position: Int) {
```
lpatel99, line 53: Perhaps fix the code climate problem by calling another function which does a part of this one.

lpatel99 (commented):
Also I recommend changing the code climate settings to allow more lines of similar code and larger files.
Also, could you add a button to view your own CV in the profile page activity?

app/src/main/java/com/sdp13epfl2021/projmag/adapter/CVAdapter.kt
arnauds5: Fixed in https://github.com/asfoury/projmag/pull/242/commits/d5dae663de9a3b4d93097f78bd7c6d187c674e6c

arnauds5:
> Also, could you add a button to view your own CV in the profile page activity?

I added your suggestion in https://github.com/asfoury/projmag/pull/242/commits/c6970b2d76ba2b83c55d9f7e169293a173400f10, as follow : 
image

giogio21 (commented):
you could maybe add a button to edit CV if we need to modify an existing

arnauds5:
> you could maybe add a button to edit CV if we need to modify an existing

There is already a button to create a CV.
If someone consider that editing a CV instead of creating a new one can be usefull, we can add it as a task for next week.

giogio21:
> > you could maybe add a button to edit CV if we need to modify an existing
> 
> There is already a button to create a CV.
> If someone consider that editing a CV instead of creating a new one can be usefull, we can add it as a task for next week.

okay

giogio21 (approved):
good for me

lpatel99 (approved):
All good, works well.

PR #245, opened 2021-05-25
feature/Notification

### LINKED ISSUE #214: Have notification when student apply for a project

giogio21:
A way to test notification is : create a new project and apply to it


app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectCreationActivity.kt
```diff
@@ -52,7 +67,18 @@ class ProjectCreationActivity : AppCompatActivity() {
...
+                Log.w("failure", "Fetching FCM registration token failed", task.exception)
+                return@OnCompleteListener
+            }
+            // Get new FCM registration token
+            val token = task.result
+            if (token != null) {
+                MyFirebaseMessagingService.token = token
+                authorToken = token
+            }
+        })
```
PaulCoral: This should be an abstraction
As mentioned in the code review, we should never have call to proprietary API
So you should create an abstraction for it (take database interface+implementation as example)

app/src/androidTest/java/com/sdp13epfl2021/projmag/ProjectEditTest.kt
```diff
@@ -22,7 +22,7 @@ class ProjectEditTest {
         val intent = Intent(context, ProjectCreationActivity::class.java)
         val project = ImmutableProject(
             "a", "b", "c", "d", "e", "f", 4,
-            listOf(), false, true, listOf(), false, "a"
+            listOf(), false, true, listOf(), false, "a",authorToken = "token"
```
lpatel99: avoid using string literals (since you are using the same value in many places, even if it is for tests). You could also define "token" as the default value for authorToken in ImmutableProject if you think this is a good idea.

app/src/main/java/com/sdp13epfl2021/projmag/activities/ProjectInformationActivity.kt
```diff
@@ -286,7 +311,7 @@ class ProjectInformationActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_project_information)
-
+        MyFirebaseMessagingService.sharedPref = getSharedPreferences("sharedPref", Context.MODE_PRIVATE)
```
lpatel99: avoid string literal

app/src/main/java/com/sdp13epfl2021/projmag/model/ImmutableProject.kt
```diff
@@ -27,7 +28,8 @@ data class ImmutableProject(
     val description: String,
     val videoUri: List<String> = listOf(),
     val allowedSections: List<String> = listOf(),
-    val creationDate: Date = Date()
+    val creationDate: Date = Date(),
+    val authorToken: String
```
lpatel99: perhaps give it a default value (as you wish)

app/src/main/java/com/sdp13epfl2021/projmag/notification/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.sdp13epfl2021.projmag.R
+import kotlin.random.Random
+
+private const val CHANNEL_ID = "my_channel"
+class MyFirebaseMessagingService: FirebaseMessagingService() {
+     companion object{
+        var sharedPref: SharedPreferences? = null
+        var token :String?
+        get(){
+            return sharedPref?.getString("token","")
```
lpatel99: avoid string literal

app/src/main/java/com/sdp13epfl2021/projmag/notification/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+        set(value){
+            sharedPref?.edit()?.putString("token",value)?.apply()
```
lpatel99: same

app/src/main/java/com/sdp13epfl2021/projmag/notification/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,74 @@
...
+        val notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+        val notificationID = Random.nextInt()
+        if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.O){
+            createNotificationChannel(notificationManager)
+        }
+        intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
+        val pendingIntent = PendingIntent.getActivity(this,0,intent,FLAG_ONE_SHOT)
+        val notification = NotificationCompat.Builder(this, CHANNEL_ID)
+            .setContentTitle(msg.data["title"])
+            .setContentText(msg.data["message"])
```
lpatel99: same

app/src/main/java/com/sdp13epfl2021/projmag/notification/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,74 @@
...
+
+
+
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private  fun createNotificationChannel(notificationManger: NotificationManager){
+        val channelName = "channelName"
+        val channel = NotificationChannel(CHANNEL_ID,channelName,IMPORTANCE_HIGH).apply {
+            description = "My channel description"
```
lpatel99: same

lpatel99 (commented):
Also, I think projects which do not have an authorToken are not coming up in the list, perhaps you can fix this by adding a default value to authorToken.
Also the notification text could read:
"New Application
A student has applied to your project: ..."
And clicking on the notification could take the user to the waiting list of their project.

app/src/main/java/com/sdp13epfl2021/projmag/notification/MyFirebaseMessagingService.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.annotation.RequiresApi
+import androidx.core.app.NotificationCompat
+import com.google.firebase.messaging.FirebaseMessagingService
+import com.google.firebase.messaging.RemoteMessage
+import com.sdp13epfl2021.projmag.MainActivity
+import com.sdp13epfl2021.projmag.R
+import kotlin.random.Random
+
+private const val CHANNEL_ID = "my_channel"
+class MyFirebaseMessagingService: FirebaseMessagingService() {
```
PaulCoral: `MyFirebaseMessagingService` is not very self explanatory. Maybe `ProjectNotificatonService` or something more related to the code that you wrote

PaulCoral (dismissed):
Also for me the use of `token` is a bit weird.
(Topic messaging could be an easy and modular way to go)

PaulCoral:
- The apply notification feature works, but it seems like it is bound to the phone and not the user (I get it on my phone and not the emulator).
- I can't get the new project notification

app/src/main/java/com/sdp13epfl2021/projmag/notification/Constants.kt
```diff
@@ -0,0 +1,10 @@
+package com.sdp13epfl2021.projmag.notification
+
+class Constants {
+    companion object{
+        const val BASE_URL = "https://fcm.googleapis.com"
+        const val SERVER_KEY= "AAAAwNSBlt8:APA91bHNLxj_wt6HCQ9Lfk3_6sSYIYHxfSRA7qgR_yZP-zCYxpgvBHlttayDPn20zKundac4-3RmyrcNwJPZagYLLuXOVOKljgVLRmIEZbLYODAXT1RblGZlddwa4U-SmCkURBlVs6TC"
+        const val  CONTENT_TYPE = "application/json"
+
+    }
+}
```
PaulCoral: You could just have written an `object Constants`.
However using an `object` is overkill here. You can just write them as **top-level** definition without nesting them inside an object.

app/src/main/java/com/sdp13epfl2021/projmag/notification/Constants.kt
```diff
@@ -0,0 +1,10 @@
+package com.sdp13epfl2021.projmag.notification
+
+class Constants {
+    companion object{
+        const val BASE_URL = "https://fcm.googleapis.com"
+        const val SERVER_KEY= "AAAAwNSBlt8:APA91bHNLxj_wt6HCQ9Lfk3_6sSYIYHxfSRA7qgR_yZP-zCYxpgvBHlttayDPn20zKundac4-3RmyrcNwJPZagYLLuXOVOKljgVLRmIEZbLYODAXT1RblGZlddwa4U-SmCkURBlVs6TC"
```
PaulCoral: # :warning:  **Attention** :warning: 

You should **never** post a **private** key on a public repository.
Anybody can use it to send notification and messages in our application, this is a serious security issue.

I suggest that you read this [documentation](https://github.com/sweng-epfl/public/blob/master/project/03-Operations/Handout.md#ignoring-files): basically using a `*.xml` file with our API keys in it, adding it to `.gitignore`, and use github secret to store the key.

Then we will certainly need to recreate a key, but we should meet with the team to be sure that we do not break anything.

PaulCoral (changes requested):
A serious issue with keys

app/src/main/java/com/sdp13epfl2021/projmag/notification/Constants.kt
```diff
@@ -0,0 +1,5 @@
+package com.sdp13epfl2021.projmag.notification
+
+const val BASE_URL = "https://fcm.googleapis.com"
+const val SERVER_KEY= "AAAAwNSBlt8:APA91bH7r-mQhoX9-KzML9N_86Hq1N6nZFO-cONdw7ghuG8SJHAmV4A3rZ-_TR8PEhxxGTGftPzThNVpYVT9_GF_a7PLobSobUdJ0pTSnUh5VvUl_p53Co-golonO3RhaXYzhUvJgSs2"
+const val  CONTENT_TYPE = "application/json"
```
lpatel99: key is still here

lpatel99 (commented):
all good except key issue and need a few tests.

app/src/main/java/com/sdp13epfl2021/projmag/notification/Constants.kt
PaulCoral: @giogio21 if you ignored the file, you had to delete it before (or just `SERVER_KEY= "" `) and then ignore it

giogio21: Normaly it is good now

PaulCoral (approved):
For me it is a good start for implementing notification. Setting up the basics.
Currently it works for notifying teacher about a student that applied to one of the project

lpatel99 (approved):

PR #246, opened 2021-05-26
Feature/comment section UI

### LINKED ISSUE #233: Have a UI element that lets student write a comment and submit the comment

asfoury:


app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -0,0 +1,24 @@
...
+    private lateinit var messageAdapter: MessageListAdapter
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_comments)
+
+        val comments = MessagesDatasource().loadMessages()
+        val recyclerView = findViewById<RecyclerView>(R.id.recycler_view_comments)
+        recyclerView.adapter = MessageListAdapter(this, comments)
+        recyclerView.setHasFixedSize(true)
```
arnauds5: Even if there is a fixed number of message now, it won't be the case with real comments
```suggestion
        recyclerView.setHasFixedSize(false)
```

app/src/main/java/com/sdp13epfl2021/projmag/model/Message.kt
```diff
@@ -0,0 +1,5 @@
+package com.sdp13epfl2021.projmag.model
+
+data class Message(val messageContent: String, val sender : ImmutableProfile, val createdAt : Long) {
```
arnauds5, line 7: This need some KDoc, in particular we can't know from here what createdAt is (timestamp/seconds/miliseconds/etc ?)

app/src/main/res/values/strings.xml
```diff
@@ -139,6 +139,12 @@
     <string name="apply_text">APPLY</string>
     <string name="unaply_text">UNAPPLY</string>
     <string name="failed">failed</string>
+    <string name="enter_comment">Enter Comment</string>
+    <string name="send_comment_icon">send_comment_icon</string>
+    <string name="march_11">March 11</string>
+    <string name="john_doe">John Doe</string>
+    <string name="hello_how_are_you">Hello! How are you?</string>
+    <string name="_1_00">1:00</string>
```
arnauds5, line 147: It seems a little strange to have some random default message/date/name for comments.

app/src/main/java/com/sdp13epfl2021/projmag/adapter/MessageListAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): MessageViewHolder {
+        val adapterLayout = LayoutInflater.from(parent.context)
+            .inflate(R.layout.message, parent, false)
+        return MessageViewHolder(adapterLayout)
+    }
+
+    override fun onBindViewHolder(holder: MessageViewHolder, position: Int) {
+        val message = messages[position]
+        holder.messageTextView.text = message.messageContent
+        holder.messageDate.text = message.createdAt.toString()
```
arnauds5: If the createdAt is the milliseconds used in Date, we can use the following code
The toString format of Date is "dow mon dd hh:mm:ss zzz yyyy",
Instead of toString, we could also use some date formater instead.
```suggestion
        holder.messageDate.text = Date(message.createdAt).toString()
```

app/src/main/java/com/sdp13epfl2021/projmag/adapter/MessageListAdapter.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageButton
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.chip.ChipGroup
+import com.sdp13epfl2021.projmag.R
+import com.sdp13epfl2021.projmag.model.Message
+
+class MessageListAdapter( private val context : Context,  private val messages : List<Message>) : RecyclerView.Adapter<MessageListAdapter.MessageViewHolder>() {
```
arnauds5, line 19: Need some basic KDoc.

arnauds5 (changes requested):

app/src/main/res/values/strings.xml
asfoury: I agree but in any case this will be deleted when we connect it to firebase so we can keep them

app/src/main/java/com/sdp13epfl2021/projmag/adapter/MessageListAdapter.kt
asfoury: Done

app/src/main/java/com/sdp13epfl2021/projmag/model/Message.kt
asfoury: I agree

app/src/androidTest/java/com/sdp13epfl2021/projmag/CommentsUITest.kt
```diff
@@ -0,0 +1,53 @@
...
+            .perform(replaceText("Hello! this is a question"))
+    }
+
+    @Test
+    fun userCanClickOnSend() {
+        onView(withId(R.id.comments_edit_text))
+            .perform(replaceText("Hello! this is a question"))
+        onView(withId(R.id.comments_send_button))
+            .perform(click())
+    }
```
PaulCoral: There's no assertion in your test.
Maybe you can add an assertion that the new message appears on the screen ?

PaulCoral (changes requested):
I am not able to send a message with the app

arnauds5:
> I am not able to send a message with the app

In this PR, there is only the UI part, nothing is done when the send button is clicked, and the messages are from a constant list (the same for all projects).

app/src/androidTest/java/com/sdp13epfl2021/projmag/CommentsUITest.kt
asfoury: Exactly like Arnaud said this is only the UI nothing is send/received yet so this is normal

arnauds5 (approved):

app/src/androidTest/java/com/sdp13epfl2021/projmag/CommentsUITest.kt
asfoury: I will add the assert

lpatel99 (approved):
Very good! All good for me except the problem pointed out by Paul.

PaulCoral (approved):
Ok for me. All issues have been addressed or explained/understood

PR #247, opened 2021-05-27
feature/Map for project creation and information

### LINKED ISSUE #231: Add map to app.

lpatel99:
With this pull request, a map is added to the application. The map can be accessed from the project creation screen, where the user creating the project can choose to add a location to the project by placing a marker on the map. When viewing project information, if the project has a location, then a map button is visible in the project information. This button takes the user to a map with a marker at the location of the project. The location is stored as doubles? latitude and longitude on Firebase.
Tests have yet to be added and thus far, the CI system does not work with this PR as it is lacking the Google Maps API key.


app/src/main/java/com/sdp13epfl2021/projmag/activities/MapsActivity.kt
```diff
@@ -0,0 +1,114 @@
...
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+
+        val nextActivity =
+            if (nextActivityString == CREATION_STRING)
+                ProjectCreationActivity::class.java
+            else
+                ProjectInformationActivity::class.java
+        val newIntent = Intent(this, nextActivity)
+        newIntent.putExtra(PROJECT_EXTRA, project)
+        if (nextActivityString == CREATION_STRING) {
```
arnauds5, line 104: Maybe doing the same as ProjectCreactionActivity/VideoSubtitlingActivity for exampe, with startActivityForResult/onActivityResult function, would be perhaps better ?
This would also move "if (fromMap) {...}" code outside to the onActivityResult function.

app/src/main/java/com/sdp13epfl2021/projmag/MainActivity.kt
```diff
@@ -22,7 +22,6 @@ class MainActivity : AppCompatActivity() {
         //These are used as name for extra in Intent
         const val fromLinkString: String = "fromLink"   //Boolean
         const val projectIdString: String = "projectId" //ProjectId
-        const val projectString: String = "project"     //ImmutableProject
```
arnauds5, line 25: Maybe better to keep it here.

arnauds5 (commented):
It seems to work correctly in my emulator, but will the apk still required to have the private key + signature key added ?

app/src/main/java/com/sdp13epfl2021/projmag/MainActivity.kt
lpatel99: I removed it because it was redundant but ok

PR #252, opened 2021-06-01
Fix/profile creation gender selector

### LINKED ISSUE #241: Gender in profile is often changed to Other if it's not the same upper/lowercase.

arnauds5:
The selection of the gender is no longer a TextEdit but a Spinner now (Selector).
The Gender enum now has a function to convert string to Gender with ignoreCase.
Fix https://github.com/asfoury/projmag/issues/241.

image 
image


PaulCoral (approved):
Good after a talk with the author

giogio21 (approved):
look goods

PR #254, opened 2021-06-02
Comment section Firebase

### LINKED ISSUE #232: Upload submitted student comments to firebase

asfoury:


app/src/main/java/com/sdp13epfl2021/projmag/adapter/ProjectAdapter.kt
```diff
@@ -142,6 +142,7 @@ class ProjectAdapter(
         holder.commentButton.setOnClickListener {
             val context = holder.view.context
             val intent = Intent(context, CommentsActivity::class.java)
+            intent.putExtra("projectId", project.id)
```
arnauds5: We should avoid string literal for extra name.
```suggestion
            intent.putExtra(MainActivity.projectIdString, project.id)
```

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/CommentsDatabase.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.sdp13epfl2021.projmag.model.Message
+
+interface CommentsDatabase {
+    /**
+     * Adds a comment message to the comments of a project
+     * Call `onSuccess` if the operation succeeded
+     * Call `onFailure` with an Exception in case of failure
+     *
+     * @param message The message to be added
+     * @param projectId The ID of the favorite project
```
arnauds5: It's maybe a typo ?
```suggestion
     * @param projectId The ID of the project
```

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/CommentsDatabase.kt
```diff
@@ -0,0 +1,49 @@
...
+        onSuccess: () -> Unit,
+        onFailure: (Exception) -> Unit
+    )
+
+    /**
+     * Gets the comments of a project
+     * Call `onSuccess` with the a `List` of messages if the operation succeeded
+     * Call `onFailure` with an Exception in case of failure
+     *
+     * @param projectId The ID of the favorite project
```
arnauds5: same ?
```suggestion
     * @param projectId The ID of the project
```

app/src/main/java/com/sdp13epfl2021/projmag/model/Message.kt
```diff
@@ -5,5 +5,36 @@ package com.sdp13epfl2021.projmag.model
...
+                else -> Success(Message(messageContent,sender,createdAt))
+            }
+        }
+    }
 
+    fun toMapString() = hashMapOf(
+        FieldNames.MESSAGE_CONTENT to messageContent,
+        FieldNames.SENDER to sender.firstName,
+        FieldNames.CREATION_DATE to createdAt
+    )
```
arnauds5: This function should have some basic KDoc.
Using the firstname only could cause problems when retrieving message.

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.sdp13epfl2021.projmag.model.*
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+class FirebaseCommentsDatabase @Inject constructor(
+    private val firestore: FirebaseFirestore,
+    private val auth: FirebaseAuth,
+    private val userdataDatabase: UserdataDatabase
+) : CommentsDatabase,  AppCompatActivity() {
```
arnauds5: This database should not be an activity.
```suggestion
) : CommentsDatabase {
```

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+            val messagesStrings = newMessages.map {message -> message.toMapString()}
+            val docData = hashMapOf(
+                "comments" to messagesStrings
+            )
+            firestore.collection(PROJECT_COMMENTS).document(projectId).set(
+                docData
+            )
+                .addOnSuccessListener{  onSuccess() }
+                .addOnFailureListener { exception -> onFailure(exception) }
+
```
arnauds5: It will be easier to read with this indentation : 
```suggestion
            firestore
                .collection(PROJECT_COMMENTS)
                .document(projectId)
                .set(docData)
                .addOnSuccessListener{ onSuccess() }
                .addOnFailureListener(onFailure)
```

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+    override fun addCommentToProjectComments(
+        message: Message,
+        projectId: ProjectId,
+        onSuccess: () -> Unit,
+        onFailure: (Exception) -> Unit
+    ) {
+        getCommentsOfProject(projectId, { oldMessages ->
+            val newMessages = oldMessages.toMutableList()
+            newMessages += message
+            val messagesStrings = newMessages.map {message -> message.toMapString()}
```
arnauds5: We could do the same, without a mutable list.
```suggestion
            val newMessages: List<Message> = oldMessages + message
            val messagesStrings = newMessages.map(Message::toMapString)
```

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    override fun getCommentsOfProject(
+        projectId: ProjectId,
+        onSuccess: (List<Message>) -> Unit,
+        onFailure: (Exception) -> Unit
+    ) {
+        val userId = auth.uid
+        if(userId != null) {
+            userdataDatabase.getProfile(userId,{ profile ->
```
arnauds5: This code will always get the profile of the current user instead of profiles of each user that sent a message.

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+                if(profile != null) {
```
arnauds5: Nothing is done if the profile is null, not even a failure callback.

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,104 @@
...
+                    firestore.collection(PROJECT_COMMENTS).document(projectId).get()
+                        .addOnSuccessListener { doc ->
+                            val messages: MutableList<Message> = mutableListOf()
+                            (doc["comments"] as? List<*>)?.forEach { comment ->
+                                val commentHash = comment as? HashMap<*, *>
+                                commentHash?.let { comment ->
+                                    messages += Message(
+                                        comment["message"] as String,
+                                        profile, comment["creationDate"] as Long
```
arnauds5: We should avoid unsafe cast (if the data type are not correct, it will throw an exception).

arnauds5 (changes requested):
There is a serious problem with how Message keep the sender.
In Message it's a ImmutableProfile, but only the firstname is stored in firebase, and the wrong profile is used when we get the messages.

I see two differents options :
- Only keep the firstname in Message (but can't be modified if the sender changed his name).
- Store userId instead of firstname and then retrieve ImmutableProfile when getting messages.

asfoury:
Their is a very bad bug that changes the sender of all the messages to the person who send the last message in the comment section of a given project will fix that tomorrow

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
asfoury: Indeed will fix that

arnauds5:
I tried to write some code that could be used if we save the userId instead of firstname to firebase (but I didn't have time to test).
If we keep profile in Message and userId on firebase, we also need to add the userId to the ImmutableProfile class.
```kotlin
    private suspend fun loadMessage(dataMap: Map<String, Any>): Message? {
        val message = dataMap["message"] as? String
        val userId = dataMap["userId"] as? String
        val createdAt = dataMap["createdAt"] as? Long
        if (message != null && userId != null && createdAt != null) {
            val profileDeferred = CompletableDeferred<ImmutableProfile?>()
            userdataDatabase.getProfile(userId, { profile ->
                profileDeferred.complete(profile)
            }, {
                profileDeferred.complete(null)
            })
            val profile = profileDeferred.await()
            if (profile != null) {
                return Message(message, profile, createdAt)
            }
        }
        return null
    }

    @Suppress("UNCHECKED_CAST")
    private suspend fun loadAllMessages(dataList: List<Map<String, Any>>): List<Message> {
        val messages = ConcurrentLinkedQueue<Message>()
        dataList.map { dataMap ->
            GlobalScope.launch(Dispatchers.IO) {
                val message = loadMessage(dataMap)
                if (message != null) {
                    messages.add(message)
                }
            }
        }.joinAll()
        return messages.toList().sortedBy(Message::createdAt)
    }

    @Suppress("UNCHECKED_CAST")
    override fun getCommentsOfProject(
        projectId: ProjectId,
        onSuccess: (List<Message>) -> Unit,
    ) {
        getDoc(projectId)
            .get()
            .addOnSuccessListener { doc ->
                val messagesList = doc["comments"] as? List<Map<String, Any>>
                messagesList?.let {
                    GlobalScope.launch {
                        onSuccess(loadAllMessages(messagesList))
                    }
                } ?: onSuccess(emptyList())
            }
            .addOnFailureListener(onFailure)
    }
```

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,103 @@
...
+    override fun addListener(
+        projectID: ProjectId,
+        onChange: (ProjectId, List<Message>) -> Unit
+    ) {
+        getCommentsOfProject(projectID, { comments ->
+            getDoc(projectID)
+                .addSnapshotListener { snapshot, _ ->
+                    snapshot?.data?.let {
+                        GlobalScope.launch {
+                            onChange(projectID, comments)
```
arnauds5: This code will always return the comments getted at the beginning, but not the new comments.

app/src/main/java/com/sdp13epfl2021/projmag/adapter/MessageListAdapter.kt
```diff
@@ -36,8 +41,13 @@ class MessageListAdapter( private val context : Context,  private val messages :
         val message = messages[position]
         holder.messageTextView.text = message.messageContent
         holder.messageDate.text = Date(message.createdAt).toString()
-        holder.messageSender.text = message.sender.firstName
-
+        // by default set it to anonymous
+        holder.messageSender.text = context.getString(R.string.sender_anonymous)
```
arnauds5: Replacing the default "John Doe" by this string in the message.xml will be better.
(and this line of code could be remove after that)

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_comments)
-
-        val comments = MessagesDatasource().loadMessages()
+        //get the users profileId
+        val profileId : String = userId
+        // send a message
+        // get the project id
+        val projectId = intent.extras?.getString("projectId")
```
arnauds5: We should try to avoid using string literal for name of extra.
```suggestion
        val projectId: String? = intent.getStringExtra(MainActivity.projectIdString)
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
+        setUpSendButton(profileId, projectId)
         val recyclerView = findViewById<RecyclerView>(R.id.recycler_view_comments)
-        recyclerView.adapter = MessageListAdapter(this, comments)
-        recyclerView.setHasFixedSize(false)
+        commentsDB.getCommentsOfProject(projectId!!, {
```
arnauds5: Instead of using !! (that throw a NPE if null), we can finish the activity if it's null, and maybe show an error message.

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
+            sendCommentButton.setOnClickListener {
+                val editTextView = findViewById<EditText>(R.id.comments_edit_text)
+                val messageText = editTextView.text.toString()
+                when (val message = Message.build(messageText, profileId, Date().time) ) {
+                    is Success ->
+                    {
+                        commentsDB.addCommentToProjectComments(message.value, (projectId ?: ""),{}, {})
+                        editTextView.text.clear()
+                    }
+                    is Failure -> Log.d("MYTEST","Error occurred when sending message : ${message.reason}")
```
arnauds5: Log/debug/etc should be avoided in the release version.
Maybe showing an error message in a Toast will be better.

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
+    private fun setUpSendButton(profileId : String?, projectId : String?) {
+        if(profileId != null) {
+            val sendCommentButton = findViewById<ImageButton>(R.id.comments_send_button)
+            sendCommentButton.setOnClickListener {
+                val editTextView = findViewById<EditText>(R.id.comments_edit_text)
+                val messageText = editTextView.text.toString()
+                when (val message = Message.build(messageText, profileId, Date().time) ) {
+                    is Success ->
+                    {
+                        commentsDB.addCommentToProjectComments(message.value, (projectId ?: ""),{}, {})
```
arnauds5: If you finish the activity if projetctId is null, this will avoid problem if null.
```suggestion
                        commentsDB.addCommentToProjectComments(message.value, projectId, {}, {})
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
+    @Named("currentUserId")
+    lateinit var userId: String
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_comments)
-
-        val comments = MessagesDatasource().loadMessages()
+        //get the users profileId
+        val profileId : String = userId
```
arnauds5: Is there a reason to use a different variable for the userId ?
Couldn't we use the userId directly ?

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,87 @@
...
+                    this.comments = it
+                    recyclerView.adapter = MessageListAdapter(commentsDB,this, comments, userDB)
+                    recyclerView.setHasFixedSize(false)
+                    recyclerView.scrollToPosition(this.comments.size - 1)
+                }, {})
+            }
+        }
+    }
+
+    private fun setUpSendButton(profileId : String?, projectId : String?) {
```
arnauds5: Using non nullable type for projectId will make more sense.
And if we use userId directly instead of profileId, we don't even need to pass it to the function call. (and will be non-null)

app/src/main/java/com/sdp13epfl2021/projmag/database/interfaces/CommentsDatabase.kt
arnauds5: Comments are only for favorite projects, so the "favorite" should be removed.

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,107 @@
...
+        onChange: (ProjectId, List<Message>) -> Unit
+    ) {
+            getDoc(projectID)
+                .addSnapshotListener { snapshot, _ ->
+                    snapshot?.data?.let {
+                        GlobalScope.launch {
+                            @Suppress("UNCHECKED_CAST")
+                            val comments = snapshot["comments"] as? List<Message>
+                            comments?.let { coms ->
+                                Log.d("MYTEST","${coms.size}")
```
arnauds5: Log/debug shouldn't stay in the release version.
```suggestion
```

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,79 @@
...
+
+            commentsDB.addListener(projectId) { _: ProjectId, _: List<Message> ->
+                this.runOnUiThread {
+                    commentsDB.getCommentsOfProject(projectId, {
+                        recyclerView.adapter = MessageListAdapter(commentsDB, this, it, userDB)
+                        recyclerView.setHasFixedSize(false)
+                        recyclerView.scrollToPosition(it.size - 1)
+                    }, {})
+                }
+            }
```
arnauds5: There is no need to call getCommentsOfProject if the listener works ?
And the code that need to be run on the Ui thread was not the call to getCommentsOfProject but its callback.
```suggestion
            commentsDB.addListener(projectId) { _: ProjectId, messages: List<Message> ->
                this.runOnUiThread {
                    recyclerView.adapter = MessageListAdapter(commentsDB, this, messages, userDB)
                    recyclerView.setHasFixedSize(false)
                    recyclerView.scrollToPosition(messages.size - 1)
                }
            }
```

app/src/main/java/com/sdp13epfl2021/projmag/model/Message.kt
```diff
@@ -4,6 +4,39 @@ package com.sdp13epfl2021.projmag.model
...
+            const val SENDER = "sender"
+            const val CREATION_DATE = "creationDate"
+        }
+        private const val MAX_MESSAGE_SIZE = 200
+        /**
+         * Function that creates a Message
+         * returns a message wrapped in a success wrapper, or a failure with the explanation wrapped as a string
+         *
+         * @param messageContent : content of the message
+         * @param sender : profile of the user that send the message
```
arnauds5: The KDoc need to be updated.
```suggestion
         * @param userId the id of the user that send the message
```

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,107 @@
...
+                                        senderId,
+                                        creationDate
+                                    )
+                                }
+                            }
+                        }
+
+                    }
+                }
+                onSuccess(messages)
```
arnauds5: This code is a little too complex to understand easily.
Instead of a cascade of ?.let, maybe a single if with all check will be more readable.
```suggestion
                val messages = (doc["comments"] as? List<*>)?.mapNotNull { comment ->
                    val commentHash = comment as? HashMap<*, *>
                     commentHash?.let {
                        val messageContent = it["message"] as? String
                        val userId = it["sender"] as? String
                        val createdAt = it["creationDate"] as? Long

                        if (messageContent != null && userId != null && createdAt != null) {
                            Message(messageContent, userId, createdAt)
                        } else {
                            null
                        }
                    }
                } ?: emptyList()
                onSuccess(messages)
```

arnauds5 (changes requested):

asfoury:
Their is a problem with the listener, I get this error with the suggested code:
image

arnauds5 (approved):
There still some minor change such as "favorite" in KDoc or default sender in the message.xml, but it looks good otherwise.

app/src/main/java/com/sdp13epfl2021/projmag/activities/CommentsActivity.kt
```diff
@@ -1,24 +1,76 @@
...
+    lateinit var userId: String
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_comments)
+        // send a message
+        // get the project id
+        val projectId: String? = intent.getStringExtra(MainActivity.projectIdString)
+        if(projectId == null){
+            finish()
```
PaulCoral, line 46: An error message would be cool

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,109 @@
...
+        message: Message,
+        projectId: ProjectId,
+        onSuccess: () -> Unit,
+        onFailure: (Exception) -> Unit
+    ) {
+        getCommentsOfProject(projectId, { oldMessages ->
+            val newMessages: List<Message> = oldMessages + message
+            val messagesStrings = newMessages.map(Message::toMapString)
+            val docData = hashMapOf(
+                "comments" to messagesStrings
```
PaulCoral, line 39: This string could be constant somewhere as it is a key

app/src/main/java/com/sdp13epfl2021/projmag/database/impl/firebase/FirebaseCommentsDatabase.kt
```diff
@@ -0,0 +1,109 @@
...
+                .addOnFailureListener(onFailure)
+
+        }, {})
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun dataListToMessages(dataList: List<*>?): List<Message> {
+        return dataList?.mapNotNull { commentData ->
+            val commentMap = commentData as? Map<String, Any>
+            commentMap?.let {
```
PaulCoral, line 55: Same here and the two other lines

PaulCoral (approved):
Fine, just check the comments

